PR #52, opened 2019-07-05
Fixed Radial Chart width and height prop

MarcusNotheis:
The props `width` and `height` were not handled correctly on the `RadialChart`.

### DIFF
--- a/packages/charts/src/components/RadialChart/index.tsx
+++ b/packages/charts/src/components/RadialChart/index.tsx
@@ -46,7 +46,7 @@ export class RadialChart extends PureComponent<RadialChartPropTypes> {
   };
 
   render() {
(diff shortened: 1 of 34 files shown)


packages/charts/src/components/RadialChart/index.tsx
```diff
@@ -66,23 +66,29 @@ export class RadialChart extends PureComponent<RadialChartPropTypes> {
       options
     );
 
+    const styles = {
```
vbersch: Maybe another name for that object or pass it inline? It shadows the styles object which gets passed to withStyles.

MarcusNotheis: is adjusted

PR #54, opened 2019-07-10
refactor: expose root html nodes via ref in every component

vbersch:
## Breaking Changes
All Components now expose their root html node/the encapsulated Web Component via ref instead of
the React instance.


packages/main/src/webComponents/Popover/demo.stories.tsx
```diff
@@ -9,6 +9,9 @@ import { PopoverVerticalAlign } from '../../lib/PopoverVerticalAlign';
 
 storiesOf('UI5 Web Components | Popover', module).add('Default', () => (
   <Popover
+    ref={(e) => {
```
MarcusNotheis: I checked in that coding by accident - I think we should remove it

PR #70, opened 2019-07-23
refactor(Page): Replace withStyles HOC by forwardRef implementation

MarcusNotheis:


packages/main/src/components/Page/index.tsx
```diff
@@ -23,89 +23,101 @@ export interface PagePropTypes extends CommonProps {
...
+  const handleNavBackButtonPress = useCallback(
+    (e) => {
+      if (typeof onNavButtonPress === 'function') {
+        onNavButtonPress(Event.of(null, e.getOriginalEvent()));
+      }
+    },
+    [onNavButtonPress]
+  );
+
+  const renderBackButton = useMemo(
```
vbersch: Please use useCallback to be more specific

packages/main/src/components/Page/index.tsx
```diff
@@ -23,89 +23,101 @@ export interface PagePropTypes extends CommonProps {
...
+        return (
+          <Button icon="navigation-left-arrow" design={ButtonDesign.Transparent} onClick={handleNavBackButtonPress} />
+        );
+      }
+      return null;
+    },
+    [showBackButton]
+  );
+
+  const renderTitle = useMemo(() => () => <Title level={TitleLevel.H5}>{title}</Title>, [title]);
```
vbersch: Please use useCallback to be more specific

packages/main/src/components/Page/index.tsx
```diff
@@ -23,89 +23,101 @@ export interface PagePropTypes extends CommonProps {
...
+
+  const renderHeader = useMemo(
```
vbersch: Whats the point in memoizing a function and calling it below? 
You could just memoize the result of the inner function and also make it dependant on renderHeader and renderCustomHeader.

vbersch: forget the dependant on renderHeader part... just dependant on renderCustomHeader of course

PR #73, opened 2019-07-25
refactor: Replace typescript loaders with babel 7

MarcusNotheis:


packages/main/src/components/ObjectPage/ObjectPageAnchorButton.tsx
```diff
@@ -19,40 +19,38 @@ interface ObjectPageAnchorPropTypes {
...
+      bottom: 0,
+      left: 0
     }
-  }),
-  {
-    name: 'ObjectPageAnchorButton'
   }
-);
+});
+const useStyles = createUseStyles<JSSTheme, keyof ReturnType<typeof anchorButtonStyles>>(anchorButtonStyles, {
```
vbersch: Is there a way to somehow abstract that typing mess? It looks pretty tedious to write in every component :/

MarcusNotheis: I tried but it did not work...
`export type JssReturnType<T> = keyof ReturnType<typeof T>` but this is not valid because is a type and a value at the same time...

PR #79, opened 2019-08-02
feat(FilterItem): Add MultiSelect option to FilterTypes

vbersch:
## Pull Request Checklist
- [ ] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents-react/blob/master/CONTRIBUTING.md)
- [ ] [Correct commit message style](https://github.com/SAP/ui5-webcomponents-react/blob/master/docs/Guidelines.md#commit-message-style)

### DIFF
--- a/packages/main/src/components/FilterItem/index.tsx
+++ b/packages/main/src/components/FilterItem/index.tsx
@@ -1,5 +1,5 @@
 import { Event, StyleClassHelper } from '@ui5/webcomponents-react-base';
-import React, { forwardRef, ReactNode, RefObject, useMemo, FC } from 'react';
+import React, { FC, forwardRef, ReactNode, RefObject, useMemo } from 'react';
 import { CommonProps } from '../../interfaces/CommonProps';
 import { BusyIndicator } from '../../lib/BusyIndicator';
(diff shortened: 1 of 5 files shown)


packages/main/src/components/FilterItem/index.tsx
```diff
@@ -52,6 +54,17 @@ const FilterItem: FC<FilterItemPropTypes> = forwardRef((props: FilterItemPropTyp
     onChange(Event.of(null, e.getOriginalEvent(), { selectedItem: item }));
   }
 
+  function onMultiCbChange(e) {
```
MarcusNotheis: would it make sense to use `useCallback` here?

vbersch: Thought about it as well... 
I think it would be a micro optimization because the only place where its used is in filterComponent which gets memoized anyway (the memoization won´t run again if onSelect or onMultiCbChange get a new reference). Therefore there is no unnecessary rerender when the reference of onMultiCbChange and onSelect changes. Thats exactly the goal of useCallback. So almost no difference with or without useCallback.

PR #80, opened 2019-08-05
fix(FilterBar): Increase margin between FilterItems in FilterBar

vbersch:


packages/main/src/components/FilterBar/FilterBar.jss.ts
```diff
@@ -28,35 +28,16 @@ const styles = ({ theme, contentDensity, parameters }: JSSTheme) => {
...
     },
-    '@media (min-width: 1024px) and (max-width: 1439px)': {
-      filterArea: {
-        columnCount: 12,
-        columnGap: '1rem',
-        rowGap: '1rem'
-      }
+    filterAreaClosed: {
+      maxHeight: '0',
+      opacity: 0
```
MarcusNotheis: I would add `padding: 0` as well, otherwise there is lots of whitespace in the bottom.

PR #83, opened 2019-08-05
fix: Re-enable passing refs to non-webcomponent components

vbersch:

### DIFF
--- a/packages/main/src/components/ObjectPageSection/index.tsx
+++ b/packages/main/src/components/ObjectPageSection/index.tsx
@@ -15,8 +15,8 @@ export interface ObjectPageSectionPropTypes extends CommonProps {
 
 const useStyles = createUseStyles<JSSTheme, keyof ReturnType<typeof styles>>(styles, { name: 'ObjectPageSection' });
 
-const ObjectPageSection: RefForwardingComponent<any, ObjectPageSectionPropTypes> = forwardRef(
-  (props: ObjectPageSectionPropTypes, ref: RefObject<any>) => {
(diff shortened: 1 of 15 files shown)


packages/main/src/components/ObjectPageSection/index.tsx
```diff
@@ -15,7 +15,7 @@ export interface ObjectPageSectionPropTypes extends CommonProps {
 
 const useStyles = createUseStyles<JSSTheme, keyof ReturnType<typeof styles>>(styles, { name: 'ObjectPageSection' });
 
-const ObjectPageSection: FC<ObjectPageSectionPropTypes> = forwardRef(
+const ObjectPageSection: RefForwardingComponent<any, ObjectPageSectionPropTypes> = forwardRef(
```
MarcusNotheis: I guess this should be a `HTMLDivElement` as well?

packages/main/src/components/VariantManagement/index.tsx
```diff
@@ -65,7 +65,7 @@ const styles = ({ parameters }: JSSTheme) => ({
 
 const useStyles = createUseStyles<JSSTheme, keyof ReturnType<typeof styles>>(styles, { name: 'Variant Management' });
 
-const VariantManagement: FC<VariantManagementPropTypes> = forwardRef(
+const VariantManagement: RefForwardingComponent<any, VariantManagementPropTypes> = forwardRef(
```
MarcusNotheis: can we be more specific than `any` here?

vbersch:
RefForwardingComponent seems to be not supported by react-docgen-typescript.

PR #89, opened 2019-08-12
refactor(ThemeProvider): Functional component

MarcusNotheis:


packages/main/src/components/ThemeProvider/index.tsx
```diff
@@ -16,47 +16,61 @@ export interface ThemeProviderProps {
...
       }
     });
-  }
+  }, [theme]);
+  const { withToastContainer, children } = props;
 
-  private static getTheme = (theme: Themes) => {
+  const isCompactSize = getCompactSize();
+
+  const contentDensity = useMemo(() => {
```
vbersch: useMemo is not really needed here as contentDensity is a string and === hits on equal strings.

packages/main/src/components/ThemeProvider/index.tsx
```diff
@@ -16,47 +16,61 @@ export interface ThemeProviderProps {
...
+    return isCompactSize ? ContentDensity.Compact : ContentDensity.Cozy;
```
vbersch: you can do that directly in themeContext useMemo.

PR #125, opened 2019-09-17
Refactor/visible props

### LINKED ISSUE #115: SegmentedButtonItem: visible props has no effect
**Describe the bug**
In the SegmentedButtonItem Component the visible props seems not have an effect.
When setting visible={false}, the Segment still will be shown

**To Reproduce**
Render Segmented Button, create 2 SegmentedButtonItem and set prop visible to false for one of them. Both will be shown.

**Expected behavior**
Only SegmentedButtonItem with visible={true}, shorthanded visible (= default) should be shown.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MAC OS
 - Browser Chrome

**Additional context**

MarcusNotheis:
Closes #115 

BREAKING CHANGE: removed visible prop, please use conditional rendering instead
BREAKING CHANGE: `SegmentedButton` and `SegmentedButtonItem` prop enabled is now renamed to disabled and logic is inverted

### DIFF
--- a/packages/main/src/components/SegmentedButton/index.tsx
+++ b/packages/main/src/components/SegmentedButton/index.tsx
@@ -1,16 +1,15 @@
 import { Event, StyleClassHelper, withStyles } from '@ui5/webcomponents-react-base';
-import React, { Children, cloneElement, Component, CSSProperties, ReactElement, RefObject } from 'react';
+import React, { Children, cloneElement, Component, CSSProperties, RefObject, ReactNode } from 'react';
 import { ClassProps } from '../../interfaces/ClassProps';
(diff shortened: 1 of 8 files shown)


packages/main/src/components/SegmentedButton/index.tsx
```diff
@@ -41,105 +43,100 @@ const styles = ({ contentDensity }) => ({
...
-  }
 
-  componentDidMount() {
-    this.updateChildElementSize();
-  }
-
-  componentDidUpdate() {
-    this.updateChildElementSize();
-  }
+    useEffect(() => {
```
vbersch: useMemo?

Issue #127, opened 2019-09-17
Fiori guidelines - Side navigation and breadcrumb


CristianDDias:
**Is your feature request related to a problem? Please describe.**
There is no Side Navigation and Breadcrumb components, such as presented in the [Fiori guidelines](https://experience.sap.com/fiori-design-web/side-navigation/#overview). Many modern applications have these components nowadays.

**Describe the solution you'd like**
I would like to have a Side Navigation and Breadcrumb components. There are these components in the Fiori Fundamentals. [Side Navigation](https://sap.github.io/fundamental/components/side-navigation.html) and [Breadcrumb](https://sap.github.io/fundamental/components/breadcrumb.html).

**Describe alternatives you've considered**
For the Side Navigation, I think about creating a component using the `List` component.
For the Breadcrumb, I think about creating a component using the `Link` component.

MarcusNotheis:
Hi @CristianDDias,

thanks for submitting this issue.
We just released the [Breadcrumbs](https://ui5-webcomponents-react.netlify.com/?path=/docs/components-breadcrumbs--default-story) in version `0.5.1-rc.14`.
The SideNavigation is currently under development, we'll let you know as soon as we are ready.

Best regards,
Marcus

CristianDDias:
Thank you @MarcusNotheis!

CristianDDias:
@MarcusNotheis, do you have any expected release date for this Side Navigation feature?

MarcusNotheis:
@CristianDDias I hope to have an initial version by end of this week

CristianDDias:
Thank you @MarcusNotheis!

MarcusNotheis:
Hi @CristianDDias,

the SideNavigation was just released in `0.5.1-rc.16` :)

PR #129, opened 2019-09-18
feat(Breadcrumbs): Initial Component Implementation

MarcusNotheis:
#127


packages/main/src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <div ref={ref} title={tooltip} style={style} className={className} slot={slot}>
+      {childrenArray.map((item, index) => {
+        if (index === childrenArray.length - 1) {
+          return item;
+        }
+        return (
+          <Fragment key={index}>
+            {item}
+            <Label style={{ margin: '0 0.25rem' }} children={SeparatorStyles[separatorStyle]} />
```
vbersch: could you extract the style object to a labelStyles variable outside of the function so that we keep the same reference on render?

packages/main/src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+          </Fragment>
+        );
+      })}
+      {currentLocationText && (
+        <>
+          <Label style={{ margin: '0 0.25rem' }}>{SeparatorStyles[separatorStyle]}</Label>
```
vbersch: same as line 43

PR #145, opened 2019-09-24
refactor(base): Deleted withStyles HOC

MarcusNotheis:
## BREAKING CHANGE
- Delete `withStyles` export from `base` package


MarcusNotheis:
In memory of the first higher order component of this project, it was

PR #185, opened 2019-10-15
feat(AnalyticalTable): offer scrollTo and scrollToItem methods on table ref

vbersch:
fix: add key to Cell component
feat: offer scrollTo and scrollToItem methods on table ref


packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -81,6 +82,7 @@ export interface TableProps extends CommonProps {
   getHeaderGroupsProps?: () => any;
   getHeaderProps?: () => any;
   getRowProps?: () => any;
+  getRowProps?: () => any;
```
MarcusNotheis: duplicate row?

PR #186, opened 2019-10-15
feat(AnalyticalTable): control row selection by prop

vbersch:
What do you prefer, selectedRowKey(e.g. row_5) or selectedRowIndex(e.g. 5)?


MarcusNotheis:
I think selectedRowIndex would be more intutive, but with the row Key we can easily select rows in a tree, so I would go for `selectedRowKey`.
Just added support for trees, but it looks like that changing the selection by mouse click in a tree has a very bad performance.

PR #192, opened 2019-10-18
Use onClick instead of onPress for button sample

cgrail:
In the readme the sample code for the button doesn't work. 

Changed in: https://github.com/SAP/ui5-webcomponents-react/pull/48


vbersch:
Good find, thank you :)

PR #209, opened 2019-10-28
fix(ObjectPage): enable scroll by dragging scrollbar

vbersch:
enable scroll by dragging scrollbar
use ResizeObserver to get notified of layout updates of ObjectPage content to recalculate filler div height

### DIFF
--- a/packages/main/src/components/ObjectPage/index.tsx
+++ b/packages/main/src/components/ObjectPage/index.tsx
@@ -26,12 +26,11 @@ import { ObjectPageAnchorButton } from './ObjectPageAnchorButton';
 import { Button } from '../../webComponents/Button';
 import { CollapsedAvatar } from './CollapsedAvatar';
 import { ObjectPageScroller } from './scroll/ObjectPageScroller';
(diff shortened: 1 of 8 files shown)


packages/main/src/components/ObjectPage/index.tsx
```diff
@@ -376,86 +377,150 @@ const ObjectPage: FC<ObjectPagePropTypes> = forwardRef((props: ObjectPagePropTyp
...
+        requestAnimationFrame(() => {
+          bindScrollEvent(contentContainer, stableContentOnScrollRef);
+          bindScrollEvent(scrollBar, stableBarOnScrollRef);
+        });
+      }
+      hideHeaderButtonPressed.current = false;
+    });
+  }, [collapsedHeader]);
+
+  useMemo(() => {
```
MarcusNotheis: wouldn't `useEffect` be the better fit here? Or do you need it before rendering?

vbersch: should be an effect, you´re right.

PR #213, opened 2019-10-31
feat(ObjectPage): Enable selecting a subSection by prop

vbersch:
selectedSubSectionId: Enable selecting a subSection by prop
scrollerRef: get a reference to the internal scroller object to be able to make the ObjectPage scroll to arbitrary dom elements


packages/main/src/components/ObjectPage/index.tsx
```diff
@@ -353,11 +357,41 @@ const ObjectPage: FC<ObjectPagePropTypes> = forwardRef((props: ObjectPagePropTyp
...
+            );
+          }
+        });
+
+        if (index) {
+          setSelectedSectionIndex(index);
+        }
+      }
+    }
+  }, [props.selectedSubSectionId, scroller.current, setSelectedSectionIndex, setSelectedSubSectionId]);
```
Lukas742: Add `mode` and `children` to dependency array

PR #223, opened 2019-11-12
fix(Notification): Removed/renamed props to adopt to UI5-Webcomponents conventions

### LINKED ISSUE #221: Notification & NotificationGroup: Unify and cleanup props
All UI5 Web Component props of type `boolean` are set to false by default and they are often prefixed with `no..`, e.g. `noIcon`, `noCloseButton`, etc. We should adopt to this convention as well.

Rename Props:
- [ ] rename prop `hideShowMoreButton` to `noShowMoreButton`
- [ ] rename prop `showCloseButton` to `noCloseButton`
- [ ] remove default prop `showCloseButton` ( will result in falsy value)

Remove Props:
- [ ] remove prop `visible`, the consumer of the API should take care of that by using conditional rendering
- [ ] remove `isChild` as this is an internal API which shouldn't be used by the consumers, change access to object access to avoid TS errors
- [ ] remove `isLastChild` as this is an internal API which shouldn't be used by the consumers, change access to object access to avoid TS errors

Lukas742:
Closes #221


packages/main/src/components/NotificationGroup/Notification.test.tsx
```diff
@@ -34,9 +34,8 @@ describe('NotificationGroup', () => {
           datetime="1337 Minutes ago"
           avatar="ME"
           priority={Priority.High}
-          visible
-          hideShowMoreButton={false}
-          showCloseButton
+          noShowMoreButton={false}
```
MarcusNotheis: can be omitted

packages/main/src/components/NotificationGroup/Notification.test.tsx
```diff
@@ -76,9 +75,8 @@ describe('NotificationGroup', () => {
...
```
MarcusNotheis: can be omitted

packages/main/src/components/NotificationGroup/Notification.test.tsx
```diff
@@ -104,9 +102,8 @@ describe('NotificationGroup', () => {
...
```
MarcusNotheis: can be omitted

packages/main/src/components/NotificationGroup/Notification.test.tsx
```diff
@@ -132,9 +129,8 @@ describe('NotificationGroup', () => {
         datetime="1337 Minutes ago"
         avatar="ME"
         priority={Priority.High}
-        visible
-        hideShowMoreButton={false}
-        showCloseButton
+        noShowMoreButton={false}
```
MarcusNotheis: can be omitted

PR #264, opened 2020-01-13
fix(AnalyticalTable): Each column can have groupable, sortable, filte…

iamuan:
fix(AnalyticalTable): Each column can have groupable, sortable, filterable options


packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -331,9 +331,9 @@ const AnalyticalTable: FC<TableProps> = forwardRef((props: TableProps, ref: Ref<
                       id={column.id}
                       {...column.getHeaderProps()}
                       isLastColumn={index === columns.length - 1}
-                      groupable={props.groupable}
-                      sortable={props.sortable}
-                      filterable={props.filterable}
+                      groupable={column.groupable !== undefined ? column.groupable : props.groupable}
```
MarcusNotheis:
```suggestion
                      groupable={column.groupable ?? props.groupable}
```

packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -331,9 +331,9 @@ const AnalyticalTable: FC<TableProps> = forwardRef((props: TableProps, ref: Ref<
...
+                      sortable={column.sortable !== undefined ? column.sortable : props.sortable}
```
MarcusNotheis:
```suggestion
                      sortable={column.sortable ?? props.sortable}
```

packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -331,9 +331,9 @@ const AnalyticalTable: FC<TableProps> = forwardRef((props: TableProps, ref: Ref<
...
+                      filterable={column.filterable !== undefined ? column.filterable : props.filterable}
```
MarcusNotheis:
```suggestion
                      filterable={column.filterable ?? props.filterable}
```

iamuan:
Hi, @MarcusNotheis.

As we know, the "undefined" usually means the property or argument is not provided by the caller.

So I thought it's better that if the property is not provided at column level, then the property at table level should be considered.

For increasing readability, how about add getProvidedArgument function like below.

```js
function getProvidedArgument(...args){
  return args.reduce((result,arg)=>{
    if(arg !== undefined && result !== undefined){
      result = arg;
    }
    return result;
  },undefined)
}

groupable={getProvidedArgument(column.groupable, props.groupable)}
sortable={getProvidedArgument(column.sortable, props.sortable)}

```

MarcusNotheis:
Hi @iamuan ,
the nullish-coalescing-operator is used for exaclty this type of issue:

> The nullish coalescing operator (??) is a logical operator that returns its right-hand side operand when its left-hand side operand is null or undefined, and otherwise returns its left-hand side operand. ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)).
So it is basically the same as the function you proposed, but JS-native and handling `null` as well. What would be the downside of using it? Do you have any objections?

iamuan:
Hi, @MarcusNotheis .

I think the "undefined" and "null" have different meanings.
When we recognize if a property is provided from caller, usually check if it is "undefined".
And in my opinion, the priority of each property should be considered by this standard.

But also It's up to your choice, I would follow that.
Please leave a comment for this, I'll fix as your decision.

Thanks.

MarcusNotheis:
Hi @iamuan ,

I see your point, but I would still go for the `nullish-coalescing` option. (We have the same pattern in some other places in the library, so if we would change it here we would have to adopt in all other places as well).
Thanks for your contribution!

PR #310, opened 2020-02-10
fix(AnalyticalTable): Show selection column with select all in header

### LINKED ISSUE #302: Analytical Table: Checkbox selection (Select All)
The `selectionMode` property for Analytical Table has a `MULTI_SELECT` option, but it involves clicking the row. Because of this, the `AnalyticalTable` does not have an option to select all the rows at once.

Use checkboxes for selection, as well as a "Select All" checkbox at the top-left of the table as can be seen here in the design guidelines. https://experience.sap.com/fiori-design-web/analytical-table-alv/
image

I've tried manually adding a column of checkboxes, as well a checkbox for the header, but it becomes very messy and requires quite a bit of custom coding that is prone to error.

**Additional context**
None.

MarcusNotheis:
For compliance with fiori specs, show an additional selection column as first column to show CheckBoxes and a CheckBox in the Header for selecting/deselecting all rows

You can hide this column by setting the `noSelectionColumn` prop on the table.

Closes #302


MarcusNotheis:
> 1. Is that empty column on the left for SINGLE_SELECT there on purpose? Is that part of the specs?

Yes, that is by design (also compare the [sap.ui.table.Table](https://ui5.sap.com/#/entity/sap.ui.table.Table/sample/sap.ui.table.sample.Selection))

> 2. The additional column doesn´t seem to play well together with the new scaleWidthMode feature. The table should be stretching for the whole available width, but it doesn´t.

I'll have look at that one

PR #315, opened 2020-02-12
fix(withWebComponent): Add handler of events those are not specified in metadata of ui5-webcomponents

gnseo:
Hi, @MarcusNotheis .

For compatibility, I've updated withWebComponent to support events those are unsupported by ui5-webcomponents since onFocus event might be supported by ui5-webcomponents in sometime, if so, a runtime error would be occurred by the change.

Please check my code if it is acceptable.


packages/main/src/internal/withWebComponent.tsx
```diff
@@ -52,8 +52,14 @@ export const withWebComponent = <T extends any>(
     return Object.keys(getWebComponentMetadata().getSlots());
   };
 
-  const getEventsFromMetadata = (): string[] => {
-    return Object.keys(getWebComponentMetadata().getEvents() || {}).filter((eventName) => !eventName.startsWith('_'));
+  const getEventsFromMetadata = (otherProps?: object): string[] => {
```
MarcusNotheis:
```suggestion
  const getEventsFromMetadata = (otherProps: = {}): string[] => {
```
You can use the default parameter value here, than you can skip the manual check in line 59

packages/main/src/internal/withWebComponent.tsx
```diff
@@ -52,8 +52,14 @@ export const withWebComponent = <T extends any>(
...
+    return Object.keys(getWebComponentMetadata().getEvents() || {})
+      .filter((eventName) => !eventName.startsWith('_'))
+      .concat(
+        Object.keys(otherProps === undefined ? {} : otherProps)
```
MarcusNotheis:
```suggestion
        Object.keys(otherProps)
```

gnseo:
Hi, @MarcusNotheis .

I've forgot to use the default value of arguments.
Thank you very much for your suggestion.

MarcusNotheis:
Hi @gnseo 
you're welcome. Unfortunately my proposal contained a syntax error. Just fixed it and will merge as soon as the CI job turn green.
Thanks again for your contribution!

PR #341, opened 2020-03-09
Expose React Table Container reference

clemai:


packages/main/src/interfaces/AnalyticalTableDomRef.ts
```diff
@@ -1,6 +1,6 @@
 import { AnalyticalTableScrollMode } from '@ui5/webcomponents-react/lib/AnalyticalTableScrollMode';
 
-// also makes use of scrollTo which is already defined somewhere in HTMLDivElement
 export interface AnalyticalTableDomRef extends HTMLDivElement {
   scrollToItem: (index: number, align?: AnalyticalTableScrollMode) => void;
+  scrollTo: (any) => void; //overrides native scrollTo function
```
MarcusNotheis:
```suggestion
  scrollTo: (scrollOffset: number) => void; //overrides native scrollTo function
```

clemai: Makes sense, but only works with "ts-ignore". Done.

PR #350, opened 2020-03-16
fix(ThemingParameters): Add temporary quick fix for missing variables

MarcusNotheis:


kleinju1017:
Theming parameters of font size are working again

PR #376, opened 2020-03-24
refactor(Event-System): Call all event handlers with a CustomEvent instance

### LINKED ISSUE #324: Adding Event Parameter Typings
Currently, we are not specifying how our parameter payload looks like for any event that we´re exposing to the User. However, it should be no problem to type those payloads at least for native react components.
Typing event parameter payloads created via `WithWebComponent` have to be explored further.

MarcusNotheis:
This PR is changing the eventing system of the library. We used to create an instance of a **custom** class called Event and wrapped the event in the following structure:
```ts
export class Event {
  readonly source: ReactNode;
  readonly originalEvent: HTMLEvent;
  readonly parameters: any;
  readonly htmlSource: EventTarget;
}
```
This API is incompatible with the UI5 Web Components API and therefore we decided to remove our custom event wrapper and use the `CustomEvent` API instead.
Now, all event handlers are called with an Object which is compatible with the `CustomEvent` API. Sometimes a UI5 Custom Event is passed, sometimes a React SyntheticEvent is passed and all details are in the `details` object.

For preventing breaking changes, I added all previous methods and a proxy for accessing the parameters object to the CustomEvent and added deprecation notices.

Closes #324

### DIFF
--- a/packages/main/src/components/AnalyticalTable/index.tsx
+++ b/packages/main/src/components/AnalyticalTable/index.tsx
@@ -1,5 +1,5 @@
 import { createComponentStyles } from '@ui5/webcomponents-react-base/lib/createComponentStyles';
-import { Event } from '@ui5/webcomponents-react-base/lib/Event';
+import { enrichEventWithDetails } from '@ui5/webcomponents-react-base/lib/Utils';
 import { StyleClassHelper } from '@ui5/webcomponents-react-base/lib/StyleClassHelper';
 import { usePassThroughHtmlProps } from '@ui5/webcomponents-react-base/lib/usePassThroughHtmlProps';
 import { TableSelectionMode } from '@ui5/webcomponents-react/lib/TableSelectionMode';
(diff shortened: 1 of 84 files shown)


packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -108,11 +108,11 @@ export interface TableProps extends CommonProps {
 
   // events
 
-  onSort?: (e?: Event) => void;
-  onGroup?: (e?: Event) => void;
-  onRowSelected?: (e?: Event) => any;
-  onRowExpandChange?: (e?: Event) => any;
-  onColumnsReordered?: (e?: Event) => void;
+  onSort?: (e: CustomEvent<{column: unknown; sortDirection: string}>) => void;
+  onGroupBy?: (e: CustomEvent<{column: unknown; isGrouped: boolean}>) => void;
```
Lukas742: This would be a breaking change. Also the `prop` name and the affiliated function has to be changed then.

MarcusNotheis: this was not intended - reverted.

vbersch:
Notification onClose event:
getParameters is not defined. It should be always there shouldn´t it?

image

MarcusNotheis:
SegmentedButton is now fixed. `getParameters` should be always there, yes. You called `getParaneters` :)

vbersch:
> SegmentedButton is now fixed. `getParameters` should be always there, yes. You called `getParaneters` :)

i feel stupid...

PR #382, opened 2020-03-25
fix(Recharts/labels): resizing of chart area if axis labels get bigger

kleinju1017:
useSetMarginLeft hook

## Pull Request Checklist
- [x] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents-react/blob/master/CONTRIBUTING.md)
- [x] [Correct commit message style](https://github.com/SAP/ui5-webcomponents-react/blob/master/docs/Guidelines.md#commit-message-style)


packages/charts/src/components/ComposedChart/index.tsx
```diff
@@ -183,7 +186,7 @@ const ComposedChart: FC<ComposedChartProps> = forwardRef((props: ComposedChartPr
       tooltip={tooltip}
       slot={slot}
     >
-      <ComposedChartLib margin={{ right: 30, top: 40, bottom: 30 }} data={dataset}>
+      <ComposedChartLib margin={{ right: 30, top: 40, bottom: 30, left: marginLeft }} data={dataset}>
```
MarcusNotheis: I would modify the `useSetMarginLeft` hook to something like `useChartMargin` and return the complete, memoized margin object

packages/charts/src/hooks/useSetMarginLeft.ts
```diff
@@ -0,0 +1,14 @@
+export const useSetMarginLeft = (dataset, labelKey) =>
```
MarcusNotheis: As said in the above comment, I would rather name this `useChartMargin` and return a full, memoized margin object here. - [ ] use `useMemo`, dependent on `dataset` and `labelKey`
- [ ] Could you split this one-call-chain into multiple calls? I think it is very hard to read.
- [ ] Instead of multiplying the length with 4, you could also make use of the `getTextWidth` method from the Utils. This will get you the exact length

PR #391, opened 2020-03-26
feat(AnalyticalTable): Add Highlight Column

### LINKED ISSUE #312: Analytical Table: Highlight Items (rows)
Currently, there is no built-in option to highlight rows as shown in the [Fiori specifications](https://experience.sap.com/fiori-design-web/analytical-table-alv/#highlight-items).

Implementation of the specifications.

I've considered doing it manually, but it is better to have a solution provided by the library, especially in the case where the specifications are modified in the future.

**Additional context**
image

MarcusNotheis:
Closes #312


MarcusNotheis:
I'm not 100% sure what is the best API for this feature.
Should we keep it as "classical" prop or should we put it somewhere in the table options?

vbersch:
I´m ok with the top level props approach. Do we even need the withRowHighlight prop? We could just infer highlighting rows as soon as hightlightField has been set.

PR #432, opened 2020-04-16
fix(AnalyticalTable): don't select row when an actionable element was clicked

### LINKED ISSUE #428: Analytical Table: Unable to click components inside cells
**Describe the bug**
I have a ui5-link and a ui5-button inside a cell in Analytical Table (Multi-Select mode). I have click event listeners for both, but the handlers are not invoked when clicking either.

**Expected behavior**
Clicking these controls should dispatch a click event.

**Actual behavior**
Clicking the controls actually select/un-select the row.

Operating System:
`@ui5/webcomponents-react` version:
Do you have a dependency to`@ui5/webcomponents` in your package.json? If yes, which version?:
0.9.0-rc.9

MarcusNotheis:
Closes #428

### DIFF
--- a/packages/main/src/components/AnalyticalTable/hooks/useSingleRowStateSelection.ts
+++ b/packages/main/src/components/AnalyticalTable/hooks/useSingleRowStateSelection.ts
@@ -18,12 +18,31 @@ const getRowProps = (rowProps, { row }) => {
   ];
 };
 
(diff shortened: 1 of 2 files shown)


MarcusNotheis:
This solution feels very cumbersome but I didn't find a better way of checking if the source element is actionable or not. Maybe https://github.com/SAP/ui5-webcomponents/issues/1489 can help here in future.

PR #535, opened 2020-05-22
chore: update to storybook 6

MarcusNotheis:


MarcusNotheis:
We might need to wait for https://github.com/storybookjs/storybook/issues/10887 before merging

MarcusNotheis:
Open Issues

- [ ] All Charts: Either disable dimensions and measures args or put the dimension and measure config into an arg as well
- [ ] All Charts: Disable all dataset args (makes so sense to change that during the demo)
- [ ] All Charts: ChartConfig should be either disabled or working as arg
- [ ] Donut/Pie Chart: Documentation for `measure` prop is wrong (doc is talking about an array but it's a single object)
- [ ] MicroBarChart: Crashing completely in docs window
- [ ] ScatterChart: Needs to be moved to the new API with args and storyName
- [ ] All Components: Props that are receiving react elements (e.g ActionSheet children, AnalyticalCard header, children, etc...) must be disabled as args. You can not write JSX into that field.
- [ ] AnalyticalTable is not accepting any args
- [ ] MessageBox story has a weird behaviour in docs mode
- [ ] BusyIndicator: what is prop sie? i think its a duplicate of the correct prop size
- [ ] Select: no args

@kleinju1017

MarcusNotheis:
Add controls to .mdx stories as well - https://github.com/storybookjs/storybook/blob/0d52f9b15e6c9fd8d6454c291d06e8d7361ed958/addons/docs/docs/props-tables.md#controls

MarcusNotheis:
## Components
- [ ] AnalyticalCard: header, children and ref are of type select with value null - those should be disabled
- [ ] Bar: contentLeft, contentMiddle, contentRight, ref must be disabled
- [ ] Form: children, ref must be disabled
- [ ] Notification: avatar must be disabled
- [ ] Page: customHeader, customFooter must be disabled
- [ ] ProgressIndicator: ref must be disabled
- [ ] SideNavigation: ref must be disabled

## Web Components
- [ ] BusyIndicator: size has no default selected value?
- [ ] Dialog: header, footer must be disabled
- [ ] Input: icon must be disabled
- [ ] List: header must be disabled
- [ ] Panel: header must be disabled
- [ ] Popover: header, footer must be disabled
- [ ] ProductSwitch: no controls at all?
- [ ] ResponsivePopover: header, footer must be disabled
- [ ] ShellBar: profile, menuItems, searchField, startButton must be disabled
- [ ] Timeline: no controls?
- [ ] Title: Children can be enabled as strings are valid children (same as button or label e.g.)
- [ ] Toast: no controls?
- [ ] ToggleButton: Children can be enabled as strings are valid children (same as button or label e.g.)

PR #541, opened 2020-05-25
refactor(i18n): respect lazy loaded translations

MarcusNotheis:


packages/base/src/hooks/useI18nBundle.ts
```diff
@@ -1,20 +1,40 @@
...
+        const next = resolveTranslations(i18nBundle, texts);
+        if (prev.length === next.length && prev.every((translation, index) => next[index] === translation)) {
+          return prev;
+        }
+        return next;
+      });
     };
     fetchAndLoadBundle();
-  }, []);
+  }, [fetchI18nBundle]);
```
Lukas742, line 37: Please add dependencies.
```suggestion
  }, [fetchI18nBundle, bundleName, texts]);
```

.storybook/preview.js
```diff
@@ -37,9 +37,9 @@ const ThemeContainer = ({ theme, contentDensity, children, direction }) => {
     }
   }, [contentDensity]);
 
-  useEffect(() => {
-    document.querySelector('html').setAttribute('dir', direction.toLowerCase());
-  }, [direction]);
+  // useEffect(() => {
+  //   document.querySelector('html').setAttribute('dir', direction.toLowerCase());
+  // }, [direction]);
```
Lukas742, line 42: Was this intentionally not removed?

MarcusNotheis: Yes, because we need some way to change the Text Direction in the storybook but this is only half of the solution. I just didn't figure out yet how to change the RTL setting of the Web Componenents

PR #574, opened 2020-06-16
docs(AnalyticalTable): add recipe on how to show less columns of smaller screens

### LINKED ISSUE #562: Analytical Table: Set Visible Columns for Different Screen Sizes
Hi,

When using an Analytical Table, is it possible to hide different columns depending on the screen size used by the user? Currently, all columns are shown regardless of the screen size.

Dekstop:

AnalyticalTableDesktop

Mobile:

AnalyticalTableMobile

It would be nice if we could display the first 2 columns on mobile devices, the first 4 on tablets and all columns on desktop, for example. 

I assume this is not currently possible since I looked at the documentation and didn't see any property in the column definition that could provide such a functionality. If it is possible, it should be added to the Analytical Table documentation and if not, could it be implemented?

Thanks!
Math

MarcusNotheis:
Closes #562

### DIFF
--- a/packages/main/src/components/AnalyticalTable/AnalyticalTable.mdx
+++ b/packages/main/src/components/AnalyticalTable/AnalyticalTable.mdx
@@ -77,5 +77,38 @@ React.useEffect(() => {
 
 For more details on this behavior you can double check the [react-table docs](https://github.com/tannerlinsley/react-table/blob/master/docs/faq.md#how-do-i-stop-my-table-state-from-automatically-resetting-when-my-data-changes).
 
+### How can I show a subset of all columns on mobile devices?
+
+
+In case you want to use the `AnalyticalTable` on mobile devices as well, it might be helpful to reduce the amount of columns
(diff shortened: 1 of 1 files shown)


packages/main/src/components/AnalyticalTable/AnalyticalTable.mdx
```diff
@@ -77,5 +77,38 @@ React.useEffect(() => {
...
+      setResponsiveColumns(columns.slice(0, 3));
+    } else {
+      setResponsiveColumns(columns);
+    }
+  }, [currentRange]);
+  return <AnalyticalTable columns={responsiveColumns} data={data} title="ResponsiveTable" />;
+};
+```
+
+With the help of that effect, the table will now show either 2 columns of a mobile phone, 3 columns of a tablet device and all columns on Desktop devices.
```
Lukas742, line 110: typo :)
```suggestion
With the help of that effect, the table will now show either 2 columns on a mobile phone, 3 columns on a tablet device and all columns on Desktop devices.
```

PR #665, opened 2020-08-21
fix(Form): fix tab order

### LINKED ISSUE #645: Form: Switch to wrong control when pressing [TAB]
Hello,

If you look at the `Form `[documentation](https://sap.github.io/ui5-webcomponents-react/master/?path=/docs/components-form--default-story), you can see that pressing [TAB] switches to the wrong control.

Currently, pressing [TAB] changes control horizontally without taking into acount the form groups. It would be a nicer experience if pressing [TAB] would set the focus on the next form item of the form group.

Thanks!
Math

MarcusNotheis:
Closes #645

### DIFF
--- a/.github/workflows/release-rc.yml
+++ b/.github/workflows/release-rc.yml
@@ -1,0 +1,45 @@
+name: 'Release RC'
+
+on: 
+  workflow_dispatch
+  
+jobs:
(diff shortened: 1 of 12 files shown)


Lukas742 (approved):
Could you remove

PR #667, opened 2020-08-25
feat(DonutChart): add option to mark chart segments as active

vbersch:


Lukas742 (approved):

packages/charts/src/components/PieChart/PieChart.tsx
```diff
@@ -143,6 +145,85 @@ const PieChart: FC<PieChartProps> = forwardRef((props: PieChartProps, ref: Ref<H
     [onDataPointClick]
   );
 
+  const renderActiveShape = useCallback(
+    (props) => {
```
MarcusNotheis, line 149: this code is copied from recharts, right? we have to add the corresponding license entry to the .reuse/dep5 file for this file

packages/charts/src/components/PieChart/PieChart.tsx
```diff
@@ -143,6 +145,85 @@ const PieChart: FC<PieChartProps> = forwardRef((props: PieChartProps, ref: Ref<H
...
+          )}
+        </g>
+      );
+    },
+    [showActiveSegmentDataLabel]
+  );
+
+  const renderLabelLine = useCallback(
+    (props) => {
+      if (chartConfig.activeSegment === props.index) return null;
```
MarcusNotheis, line 205: what about the previous check on `measure.hideDataLabel`?

MarcusNotheis (changes requested):

packages/charts/src/components/PieChart/PieChart.tsx
MarcusNotheis: compare e.g with the web components project: https://github.com/SAP/ui5-webcomponents/blob/master/.reuse/dep5

MarcusNotheis (approved):

PR #676, opened 2020-09-10
fix(AnalyticalTable): control row-selection explicitly by setting selectRowIds

Lukas742:
BREAKING CHANGE: The row-selection has now manually be handled if `selectedRowIds` is set. If you want a row initially selected, but still want the table to handle row-selection, please use `reactTableOptions={{ initialState: { selectedRowIds: { [key: string]: boolean } } }}`

In my opinion we need to change the behavior of `selectedRowIds`. If the user sets this prop, the table's row selection should then be controlled manually. This will prevent cases where the table's selection logic breaks, because of double fired `toggleRowSelected` and also is more consistent to how the other react-table props are implemented. If the user wants to select a row initially, it's still possible by setting the initialState.

What do you think? @vbersch @MarcusNotheis


Lukas742:
closed in favor of https://github.com/SAP/ui5-webcomponents-react/pull/685

PR #686, opened 2020-09-24
feat: support server side rendering

### LINKED ISSUE #194: Support for Static Site Generators
I really like UI5 WebComponents for React. I wanted to use it in combination with a static site generator like Next.js or Gatsby.js, but it didn't work. I don't think Server Side Rendering is supported currently. Would be great if you could support that to improve the speed of the initial page loading.

MarcusNotheis:
Closes #194


MarcusNotheis:
This approach is not feasible, closing this PR.

Issue #704, opened 2020-10-05
AnalyticalTable: add headerTooltip property to column


MarcusNotheis:
We have the requirement to show a different tooltip for the Table Column Header than displayed as column header.
The end-user API should look like:
```jsx
<AnalyticalTable 
   columns={[
     {
        Header: 'This is my Column Title',
+       headerTooltip: 'this is a more extensive description....'
      }
   , {}, ...]}
    ...
   /> 
```

## Todos:

- [ ] add the `headerTooltip: string` property to the [`AnalyticalTableColumnDefinition`](https://github.com/SAP/ui5-webcomponents-react/blob/master/packages/main/src/interfaces/AnalyticalTableColumnDefinition.ts) interface
- [ ] read the new `headerTooltip` property from `column` in the [`ColumnHeaderContainer`](https://github.com/SAP/ui5-webcomponents-react/blob/master/packages/main/src/components/AnalyticalTable/ColumnHeader/ColumnHeaderContainer.tsx#L90) and pass it into the `ColumnHeader`.
- [ ] In the [`ColumnHeader`](https://github.com/SAP/ui5-webcomponents-react/blob/master/packages/main/src/components/AnalyticalTable/ColumnHeader/index.tsx#L180), read the new tooltip property and use is as tooltip for the `Text`. Please make sure to fallback to `children` in case the `headerTooltip` property is not defined.
- [ ] Don't forget to add the new property to the [documentation](https://github.com/SAP/ui5-webcomponents-react/blob/master/packages/main/src/components/AnalyticalTable/AnalyticalTable.stories.mdx) (section `Optional Properties`)
- [ ] as a plus: code coverage via unit tests are always welcome :)

whyang0808:
Hi @MarcusNotheis , may I pick this up?

MarcusNotheis:
Hey @whyang0808,
sure, I've just assigned this issue to you.
Thanks for picking it up! :)

PR #721, opened 2020-10-09
feat: add header tooltip property to column

whyang0808:
# Issue #704  

### Changes
- feat: add header tooltip property for the Analytical Table Column Header


MarcusNotheis (approved):
Looks good to me

PR #997, opened 2020-11-23
feat(AnalyticalTable): add subcomponents

Lukas742:
- Add new prop `renderRowSubComponent`

  - Expects a function with an optional row parameter that contains the `row` instance.
  - Can be controlled by returning different subcomponents based on the row id.
  - If the function returns `false`, `undefined` or `null` no chevron will be displayed for the corresponding row.


packages/main/src/components/AnalyticalTable/AnalyticalTable.test.tsx
```diff
@@ -372,5 +372,35 @@ describe('AnalyticalTable', () => {
     expect(row).toHaveAttribute('data-is-selected');
   });
 
+  test('render subcomponets', () => {
```
MarcusNotheis, line 375:
```suggestion
  test('render subcomponents', () => {
```

packages/main/src/components/AnalyticalTable/TableBody/VirtualTableBody.tsx
```diff
@@ -37,17 +38,28 @@ export const VirtualTableBody = (props: VirtualTableBodyProps) => {
...
 
   const consolidatedParentRef = useConsolidatedRef(parentRef);
   const rowVirtualizer = useVirtual({
     size: itemCount,
     parentRef: consolidatedParentRef,
-    estimateSize: React.useCallback(() => internalRowHeight, [internalRowHeight]),
+    estimateSize: React.useCallback(
+      (index) => {
+        if (renderRowSubComponent && rows[index].isExpanded && rowSubComponentsHeight.current?.[index]) {
+          return internalRowHeight + rowSubComponentsHeight.current?.[index];
```
MarcusNotheis, line 57:
```suggestion
          return internalRowHeight + rowSubComponentsHeight.current?.[index] ?? 0;
```
Otherwise it would be number + undefined = NaN

packages/main/src/components/AnalyticalTable/TableBody/VirtualTableBody.tsx
```diff
@@ -37,17 +38,28 @@ export const VirtualTableBody = (props: VirtualTableBodyProps) => {
...
   const overscan = overscanCount ? overscanCount : Math.floor(visibleRows / 2);
 
   const consolidatedParentRef = useConsolidatedRef(parentRef);
   const rowVirtualizer = useVirtual({
     size: itemCount,
     parentRef: consolidatedParentRef,
-    estimateSize: React.useCallback(() => internalRowHeight, [internalRowHeight]),
+    estimateSize: React.useCallback(
+      (index) => {
+        if (renderRowSubComponent && rows[index].isExpanded && rowSubComponentsHeight.current?.[index]) {
```
MarcusNotheis, line 56:
```suggestion
        if (renderRowSubComponent && rows[index].isExpanded && rowSubComponentsHeight.current.hasOwnProperty(index)) {
```

packages/main/src/components/AnalyticalTable/TableBody/VirtualTableBody.tsx
```diff
@@ -37,17 +38,28 @@ export const VirtualTableBody = (props: VirtualTableBodyProps) => {
...
+          return internalRowHeight + rowSubComponentsHeight.current?.[index];
+        }
+        return internalRowHeight;
+      },
+      [internalRowHeight, rows, renderRowSubComponent, rowSubComponentsHeight.current]
```
MarcusNotheis, line 61: I think you don't need that dependency here, the object in the ref is never recreated so it always stays the same

MarcusNotheis (changes requested):
Some minor things as comment.
And kudos for the test :)

MarcusNotheis (approved):

PR #1044, opened 2020-11-30
feat(AnalyticalTable): add visibleRowCountMode prop that controls how the table displays visible rows

Lukas742:


packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -90,12 +93,21 @@ export interface TableProps extends Omit<CommonProps, 'title'> {
    */
   minRows?: number;
   /**
-   * The number of rows visible without going into overflow.<br />
+   * Defines how the table table will render visible rows.
+   *
+   * - __"Fixed":__ The table always has as many rows as defined in the `visibleRowCount` prop.
+   * - __"Auto":__ The table automatically fills the height of the surrounding container.
```
MarcusNotheis, line 99: Maybe we should mention that we recommend to have a fixed height for the outer container?

MarcusNotheis (approved):

PR #1143, opened 2021-01-08
feat(AnalyticalTable): add onRowClickEvent

### LINKED ISSUE #1138: [AnalyticalTable] Add onRowClick prop
Our project requires the following behavior for the AnalyticalTable:
1- Clicking on the checkbox of the row selects it (standard behavior using `selectionBehavior={TableSelectionBehavior.RowSelector}`, like this: https://codesandbox.io/s/inspiring-lamarr-hx416?file=/src/App.js)
2- Clicking anywhere else on the row should trigger an event so that we can do some actions on our side (like open a dialog, side-menu, etc..) **but it shouldn’t select the row**.

Currently there's no way to have this row selection alongside a row click behavior.

Adding a `onRowClick` prop would allow us to execute actions when clicking on a row and to support row selection when clicking on the checkbox.

MarcusNotheis:
closes #1138

### DIFF
--- a/packages/main/src/components/AnalyticalTable/AnalyticalTable.stories.mdx
+++ b/packages/main/src/components/AnalyticalTable/AnalyticalTable.stories.mdx
@@ -21,7 +21,7 @@ import { TextAlign } from '@ui5/webcomponents-react/lib/TextAlign';
 import { DefaultLoadingComponent } from './defaults/LoadingComponent';
 import { DefaultNoDataComponent } from './defaults/NoDataComponent';
 import { DocsCommonProps } from '@shared/stories/DocsCommonProps';
-import { useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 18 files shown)


Lukas742 (commented):
When the `selectionMode` is set to `None`, `onRowClick` is not called. Is this intended? 
In my opinion, it makes sense to allow this in even this mode, since the callback itself has nothing to do with the selection. What do you think?

Other than that, it looks good to me.

MarcusNotheis:
> When the `selectionMode` is set to `None`, `onRowClick` is not called. Is this intended?
> In my opinion, it makes sense to allow this in even this mode, since the callback itself has nothing to do with the selection. What do you think?
> 
> Other than that, it looks good to me.

yes you're right - changed

Lukas742 (approved):

PR #1147, opened 2021-01-11
fix(ThemeProvider): remove deprecated rtl and contentDensity prop

Lukas742:
BREAKING CHANGE: __ThemeProvider:__ remove deprecated `theme` property. To check the current theme you can use:  `import { getTheme } from '@ui5/webcomponents-base/dist/config/Theme';`
BREAKING CHANGE: __ThemeProvider:__ remove deprecated `contentDensity` property from `theme`. To check the `contentDensity` you can search for the corresponding class. (e.g.:`document.body.classList.contains('ui5-content-density-compact');`
BREAKING CHANGE: __ThemeProvider:__ remove deprecated `rtl` property from `theme`

---

BREAKING CHANGE: The `ThemeProvider` context is now simplified to provide only object with `ThemingParameters`, all other properties have been removed. For details on the migration check out our [migration guide](https://sap.github.io/ui5-webcomponents-react/master/?path=/docs/migration-guide--page#themeprovider-context-changes).


MarcusNotheis:
I would go even one step further and simplify the `ThemeContext` even more: 
```
return <ReactJssThemeProvider theme={ThemingParameters}>{children}</ReactJssThemeProvider>;
```
Our `theme` is not attached to any theme change, so this value will be stale after having the Theme changed for the first time.

PR #1196, opened 2021-01-21
[Do Not Merge] chore: drop support for Edge Legacy - remove Edge Legacy polyfill

Lukas742:
BREAKING CHANGE: The Legacy Edge browser is no longer supported, therefore the following import has been removed: `import '@ui5/webcomponents-react-base/polyfill/Edge';` If your app needs to support Edge Legacy, treat it as IE11 and transpile the code to ES5 as you would for IE11.


MarcusNotheis (approved):
Looks good, but please don't merge this PR yet. I would prefer to wait until the new web components rc is available.

PR #1252, opened 2021-02-03
[do not merge] chore: deprecate Internet Explorer 11

MarcusNotheis:
In alignment with `UI5 Web Components` (https://github.com/SAP/ui5-webcomponents/pull/2767), `UI5 Web Components for React` will no longer officially support Internet Explorer 11 starting with `0.14.0`.
We will not remove any IE 11 specific code as of now, but we are officially deprecating the use of Internet Explorer 11 with our components and we might remove such code in future without further reference. In addition to that, we will not apply bug fixes for IE11 anymore.

BREAKING CHANGE: the use of our components in Internet Explorer 11 is now deprecated


packages/base/src/polyfill/IE11.ts
```diff
@@ -1,3 +1,24 @@
+/**
+ * PLEASE DO NOT IMPORT THIS FILE!
+ * IT IS ONLY EXISTING FOR NOT BREAKING LEGACY APPLICATIONS BUT OFFICIALLY DEPRECATED AND MAY BE REMOVED WITHOUT FURTHER NOTICE.
```
Lukas742, line 3:
```suggestion
 * IT IS ONLY EXISTING FOR NOT BREAKING LEGACY APPLICATIONS, BUT IS OFFICIALLY DEPRECATED AND MAY BE REMOVED WITHOUT FURTHER NOTICE.
```

README.md
```diff
@@ -175,22 +175,14 @@ For Browser Support and the configuration of the UI5 Web Components, please take
...
-
-In case you are using [`browserslist`](https://github.com/browserslist/browserslist) (which is part of `create-react-app`), please also make sure you have added `IE 11` to your browserslist configuration (can be found in `package.json` if you are using CRA).
+UI5 Web Components are supported by all major modern browsers.
+
+Browser | Supported versions
+--------|---------------------
+Chrome  | Latest
+Firefox | Latest
+Safari  | Latest
+Edge    | Latest
```
Lukas742, line 185: Should we also include that we support all major mobile browsers?

```suggestion
Edge    | Latest
Opera   | Latest
```

Lukas742 (changes requested):

README.md
MarcusNotheis: I'm not sure about this one to be honest - I've just copied the list from the `@ui5/webcomponents` project to be honest.
Which mobile browsers would you see in this list? Because with Safari and Chrome we should cover most of the iOS/Android devices, right?

Lukas742: I think it would be sufficient if we mention that we also include modern mobile browser without adding a browser list.

Or we could remove the list completely, since it should be apparent which browsers are meant.

Lukas742 (approved):

PR #1281, opened 2021-02-10
MicroBarChart - Integrade custom maxvalue

ArneBuetergerds:
In the MircoBarChart component the value bar is limited by the max dataset value.

To customize the max value of the bar I integrade the optional property customMaxValue as an number.

To ensure that the values of the dataset will be shown correctly the customMaxValue has to be higher as the maxDatasetValue.


packages/charts/src/components/MicroBarChart/MicroBarChart.tsx
```diff
@@ -61,6 +61,10 @@ export interface MicroBarChartProps
    * - `DataLabel`: a custom component to be used for the data label
    */
   measure: MeasureConfig;
+  /**
+   * An optional number for the maxValue of the valueBar
+  */
+  customMaxValue?: number;
```
Lukas742, line 67: Please rename the prop to `maxValue` as this explains more that the user is in full control over the max value of the `MicroBarChart`.

packages/charts/src/components/MicroBarChart/MicroBarChart.tsx
```diff
@@ -61,6 +61,10 @@ export interface MicroBarChartProps
    * - `DataLabel`: a custom component to be used for the data label
    */
   measure: MeasureConfig;
+  /**
+   * An optional number for the maxValue of the valueBar
+  */
```
Lukas742, line 66: Please also mention that if this prop is not set, the maximum value is equal to the highest number of the corresponding accessor in the dataset.

packages/charts/src/components/MicroBarChart/MicroBarChart.tsx
```diff
@@ -144,7 +148,13 @@ const MicroBarChart: FC<MicroBarChartProps> = forwardRef((props: MicroBarChartPr
     [props.measure]
   );
 
-  const maxValue = dataset ? Math.max(...dataset?.map((item) => getValueByDataKey(item, measure.accessor))) : 0;
+  const maxValue = useMemo(() => {
+    if (dataset) {
+      const maxDatasetValue = Math.max(...dataset?.map((item) => getValueByDataKey(item, measure.accessor)));
+      return customMaxValue && customMaxValue > maxDatasetValue ? customMaxValue : maxDatasetValue;
```
Lukas742, line 154: As the user should be in full control over the `maxValue` prop, this can be simplified:
```js
      return props.maxValue ?? maxDatasetValue;
```

Lukas742 (changes requested):
Hi @ArneBuetergerds 

thanks for creating this PR. :) 

I would even take it a step further and let the user fully control the `maxValue` when the prop is set. Therefore, it should also be possible to pass a number that is smaller than the maximum value of the data set.

ArneBuetergerds:
HI @Lukas742.

You are right. I changed the code with an new commit.

Lukas742:
@ArneBuetergerds looks good, thanks.

Could you sign the CLI? After that we are good to merge.

ArneBuetergerds:
@Lukas742, i finally figured out my problem with the cla. Now it`s signed.

Lukas742 (approved):

Lukas742:
Hi @ArneBuetergerds 

thanks for the contribution, the changes will be available with our next release :)

PR #1503, opened 2021-04-13
fix(Breadcrumbs): add missing aria labels & allow navigation with arrow keys

Lukas742:
- Add missing aria labels
- Use unordered list to display items
- Only focus first or previously focused item with tab
- Allow right and left navigation between breadcrumbs with arrow keys

### DIFF
--- a/packages/main/src/components/Breadcrumbs/index.tsx
+++ b/packages/main/src/components/Breadcrumbs/index.tsx
@@ -13,7 +13,7 @@ import React, {
   Ref,
   useCallback,
   useRef,
-  useState
+  useReducer
 } from 'react';
 import { createUseStyles } from 'react-jss';
 import { CommonProps } from '../../interfaces/CommonProps';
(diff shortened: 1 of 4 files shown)


packages/main/src/components/Breadcrumbs/index.tsx
```diff
@@ -30,35 +42,127 @@ export interface BreadcrumbsPropTypes extends CommonProps {
...
+    currentLocation: {
+      fontFamily: ThemingParameters.sapFontFamily,
+      fontSize: ThemingParameters.sapFontSize,
+      color: ThemingParameters.sapContent_LabelColor,
+      whiteSpace: 'nowrap',
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+      '&:focus': {
+        outline: `${ThemingParameters.sapContent_FocusWidth} ${ThemingParameters.sapContent_FocusStyle} ${ThemingParameters.sapContent_FocusColor}`,
+        outlineOffset: '-0.0625rem'
```
MarcusNotheis: shouldn't the offset be the same as the outline width? 
```suggestion
        outlineOffset: ThemingParameters.sapContent_FocusWidth
```

MarcusNotheis (approved):

PR #1545, opened 2021-05-03
fix: improve typings for event handlers

MarcusNotheis:
Previously our typing for all event handlers were using `CustomEvent` which has no option to add details about the current Event Target, so access `event.target.value` for the `Input - onChange` event was not possible in TypeScript strict mode.
With this we are providing our own custom event which allows to specify details about the current event target.

Fixes #1142


packages/main/src/webComponents/MultiInput/index.tsx
```diff
@@ -1,13 +1,14 @@
...
 import { withWebComponent, WithWebComponentPropTypes } from '@ui5/webcomponents-react/dist/withWebComponent';
+import { Ui5CustomEvent } from '@ui5/webcomponents-react/interfaces/Ui5CustomEvent';
 import { FC, ReactNode } from 'react';
 
 import '@ui5/webcomponents/dist/MultiInput';
 
 export interface MultiInputPropTypes extends Omit<WithWebComponentPropTypes, 'onChange' | 'onInput'> {
   /**
-   * Determines whether a value help icon will be should in the end of the input. Pressing the icon will fire `onValueHelpTrigger` event.
+   * Determines whether a value help icon will be should in the end of the input. Pressing the icon will fire `value-help-trigger` event.
```
Lukas742, line 11:
```suggestion
   * Determines whether a value help icon will be should in the end of the input. Pressing the icon will fire `onValueHelpTrigger` event.
```

packages/main/src/webComponents/MultiInput/index.tsx
```diff
@@ -146,38 +147,45 @@ export interface MultiInputPropTypes extends Omit<WithWebComponentPropTypes, 'on
...
 /**
+ * ### Overview
+ *
  * A `MultiInput` field allows the user to enter multiple values, which are displayed as `Token`. User can choose interaction for creating tokens. Fiori Guidelines say that user should create tokens when:
  *
- * *   Type a value in the input and press enter or focus out the input field (`onChange` event is fired)
- * *   Select a value from the suggestion list (`onSuggestionItemSelect` event is fired)
+ * *   Type a value in the input and press enter or focus out the input field (`change` event is fired)
+ * *   Select a value from the suggestion list
+ * (`suggestion-item-select` event is fired)
```
Lukas742, line 188:
```suggestion
 * *   Type a value in the input and press enter or focus out the input field (`onChange` event is fired)
 * *   Select a value from the suggestion list (`onSuggestionItemSelect` event is fired)
```

packages/main/src/webComponents/ShellBarItem/index.tsx
```diff
@@ -22,10 +23,8 @@ export interface ShellBarItemPropTypes extends WithWebComponentPropTypes {
   text?: string;
   /**
    * Fired, when the item is pressed.
-   *
-   * __Note:__ To get the DOM reference of the clicked `ShellBarItem`, use `event.detail.targetRef`.
```
Lukas742, line 26: Was this info intentionally removed?

packages/main/src/webComponents/Table/index.tsx
```diff
@@ -39,7 +40,7 @@ export interface TablePropTypes extends WithWebComponentPropTypes {
    */
   noDataText?: string;
   /**
-   * Defines if the value of `noDataText` will be displayed when there is no rows present in the table.
+   * Defines if the value of `noDataText` will be diplayed when there is no rows present in the table.
```
Lukas742, line 43:
```suggestion
   * Defines if the value of `noDataText` will be displayed when there is no rows present in the table.
```

packages/main/src/webComponents/UploadCollection/index.tsx
```diff
@@ -42,17 +43,17 @@ export interface UploadCollectionPropTypes extends Omit<WithWebComponentPropType
    *
    * **Note:** The `drop` event is fired only when elements are dropped within the drag and drop overlay and ignored for the other parts of the `UploadCollection`.
    */
-  onDrop?: (event: CustomEvent<{ dataTransfer: DataTransfer }>) => void;
+  onDrop?: DragEventHandler<HTMLElement>;
```
Lukas742, line 46: Does `dataTransfer` not exist anymore, or was this forgotten with the refactoring?

Lukas742 (changes requested):

packages/main/src/webComponents/ShellBarItem/index.tsx
MarcusNotheis: No :D

packages/main/src/webComponents/UploadCollection/index.tsx
MarcusNotheis: `dataTransfer` is defined by the `DragEventHandler`, having that as part of the `detail` is actually wrong

Lukas742 (approved):

PR #1689, opened 2021-06-07
fix(Form): align CheckBox, RadioButton and Switch correctly

Lukas742:


packages/main/src/components/FormItem/index.tsx
```diff
@@ -99,9 +121,21 @@ const FormItem: FC<FormItemProps> = (props: FormItemProps) => {
     );
   }
 
+  const inlineLabelStyles = useMemo(() => {
```
MarcusNotheis, line 124: no need for `useMemo`

packages/main/src/components/FormItem/index.tsx
```diff
@@ -16,13 +26,25 @@ export interface FormItemProps {
   children: ReactNode | ReactNodeArray;
 }
 
+//TODO: remove this when ui5-webcomponents adjusted this in their repo
+addCustomCSS(
+  'ui5-checkbox',
```
MarcusNotheis, line 31: do we really want to add that for all checkboxes?

MarcusNotheis (changes requested):

MarcusNotheis (approved):

PR #1706, opened 2021-06-09
fix(AnalyticalTable): respect subComponent height when visibleRowCountMode=Auto

texttechne:
This PR fixes one specific problem:
The subComponent height is respected when calculating the height of the table container when visibleRowCountMode is set to AUTO. This problem occurs if the table only renders few results.

I've added a story to demonstrate that it works which should be deleted before merging.


Lukas742:
Hi @texttechne 

thanks a lot for creating this PR! 

Unfortunately there are some issues that have to be resolved before we are able to merge this. 

- The `tableBodyHeight` is increased when expanding subcomponents, but never decreased.

- With a dataset of more entries than visible rows (e.g. `generateData(10)`, `visibleRows={3}`) when expanding subcomponents or setting `alywaysShowSubcomponent` to true, both the table body and the whole table container are overflowing, causing two scrollbars.

- In "Fixed" and "Interactive" mode expanding a subcomponent should not increase the table body container beyond the `visibleRowCount` height. It should behave similar to the tree-table.

I'm sorry that I cannot go into more detail, but I'm currently pretty busy. If you have any feedback or questions, please don't hesitate to ask.

Lukas742:
Closed due to inactivity.

PR #1768, opened 2021-06-28
fix(AnalyticalTable): Only use `offsetTop` if parent is relative positioned

Lukas742:


packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -486,8 +486,12 @@ const AnalyticalTable: FC<TableProps> = forwardRef((props: TableProps, ref: Ref<
 
   const updateRowsCount = useCallback(() => {
     if (visibleRowCountMode === TableVisibleRowCountMode.AUTO && analyticalTableRef.current?.parentElement) {
-      const tableYPosition = analyticalTableRef.current?.offsetTop ?? 0;
-      const parentHeight = analyticalTableRef.current?.parentElement?.getBoundingClientRect().height;
+      const parentElement = analyticalTableRef.current?.parentElement;
+      const tableYPosition =
+        parentElement?.style.position === 'relative' && analyticalTableRef.current?.offsetTop
```
MarcusNotheis, line 491:
```suggestion
        parentElement && getComputedStyle(parentElement).position === 'relative' && analyticalTableRef.current?.offsetTop
```
I think using `getComputedStyle` is the cleaner solution here

MarcusNotheis (changes requested):

MarcusNotheis (approved):

PR #1814, opened 2021-07-12
chore: use UI5 Web Component extension for ObjectPageAnchorTabs

MarcusNotheis:


packages/main/src/components/ObjectPage/ObjectPageAnchorTab.ts
```diff
@@ -0,0 +1,141 @@
...
+    'show-sub-sections': {}
+  }
+};
+
+class ObjectPageAnchorTab extends Tab {
+  static get metadata() {
+    return metadata;
+  }
+
+  static async onDefine() {
```
vladitasev, line 42: This is deprecated, the new one is:
```js
static get dependencies() {
  return [Icon];
}
```

and then you don't override `onDefine` but let it read the dependencies internally.

packages/main/src/components/ObjectPage/ObjectPageAnchorTab.ts
```diff
@@ -0,0 +1,141 @@
...
+    await Icon.define();
+  }
+
+  static get stripTemplate() {
+    return (context, tags, suffix) => {
+      setTags(tags);
+      setSuffix(suffix);
```
vladitasev, line 49: Probably not a concern right now, but this code will not be affected by suffixing since internally used components (`ui5-icon`) are hard-coded.

The litHTML unsafe static function must be used to wrap all calls to `ui5-` tags in the template.

vladitasev (approved):

Lukas742 (approved):

PR #1824, opened 2021-07-14
feat(ColumnChartWithTrend): initial component implementation

### LINKED ISSUE #1733: Show second y axis of charts above the primary one
It should be possible to show the second y axis above the primary one in the composed chart, so the axis are not overlapping. There should be a possibility to change the visualization of the chart like in the SAC (e.g a line chart is shown above a bar chart).

Bildschirmfoto 2021-06-15 um 12 04 41

So the line chart in the example above should be separated from the bar chart and the second y axis should be placed above the axis of the bar chart.

kleinju1017:
- added new Chart to render a Column Chart with corresponding trend line above
- expose syncId to synchronize multiple charts

Closes #1733


packages/charts/src/components/ColumnChartWithTrend/ColumnChartWithTrend.stories.mdx
```diff
@@ -0,0 +1,75 @@
...
+          formatter: (val) => val.toLocaleString(),
+          type: 'line'
+        },
+        {
+          accessor: 'sessions',
+          label: 'Active Sessions',
+          type: 'column'
+        }
+      ]
+    }}
```
Lukas742, line 49: Please move default `args` to `Meta` component, and only overwrite story-specific `args` on the `Story` itself. This saves a lot of time when creating or changing stories :)

packages/charts/src/components/ColumnChartWithTrend/ColumnWithTrendChart.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+import * as React from 'react';
+import { complexDataSet } from '../../resources/DemoProps';
+import { ColumnWithTrendChart } from './ColumnWithTrendChart';
+
+describe('ColumnChart', () => {
+  test('Renders with data', () => {
+    const utils = render(
+      <ColumnWithTrendChart
+        dataset={complexDataSet}
+        style={{ width: '50%' }}
```
Lukas742, line 11: not necessary for tests unless you want to test if the style is passed correctly. Also this will have no effect on the actual component since `JSDOM` has no mocked width or height, it is only reflected in the snapshot.

packages/charts/src/components/ColumnChartWithTrend/ColumnWithTrendChart.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+          }
+        ]}
+      />
+    );
+    expect(utils.asFragment()).toMatchSnapshot();
+  });
+
+  test('Loading placeholder', () => {
+    const wrapper = render(<ColumnWithTrendChart style={{ width: '30%' }} dimensions={[]} measures={[]} />);
+    expect(wrapper.asFragment()).toMatchSnapshot();
```
Lukas742, line 44: Please don't rely on snapshot testing only. Here you could for example check if there are two "Loading..." tooltips, one for each chart.
```
    const { asFragment, getAllByText } = render(
      <ColumnWithTrendChart style={{ width: '30%' }} dimensions={[]} measures={[]} />
    );
    const titles = getAllByText('Loading...');
    expect(titles).toHaveLength(2);
    expect(asFragment()).toMatchSnapshot();
    ```

**Review by Lukas742 — CHANGES_REQUESTED** [review]
- Please add a few more tests that don't rely on snapshot testing only. E.g. click handling, hover, etc. (Of cause only if possible ;) )

- The `ColumnChartWithTrendProps` interface extends `IChartBaseProps`, but many of them are not used at all. E.g. click handling (`onClick`, `onDataPointClick`, etc.) is not possible. If a prop is not feasible to implement with this Chart please omit it with `Omit<CommonProps, 'propToOmit'>`

File: packages/charts/src/components/ColumnChartWithTrend/Placeholder.tsx [hunk 4]
```diff
@@ -0,0 +1,32 @@
+import { ThemingParameters } from '@ui5/webcomponents-react-base/dist/ThemingParameters';
+import React from 'react';
+import ContentLoader from 'react-content-loader';
+
+export const ColumnWithTrendChartPlaceholder = (props) => {
```
Lukas742, line 5: Please rename to `ColumnChartWithTrendPlaceholder`.

Lukas742 (changes requested):

packages/charts/src/components/ColumnChartWithTrend/ColumnChartWithTrend.stories.mdx
```diff
@@ -0,0 +1,71 @@
...
+</Canvas>
+
+### Loading Placeholder
+
+<Canvas>
+  <Story
+    name="Loading Placeholder"
+    args={{
+      dataset: [],
+      style: { height: '60vh' },
```
MarcusNotheis, line 58: you can omit this line, will be handled by the default args then

MarcusNotheis (changes requested):

MarcusNotheis (approved):

PR #1831, opened 2021-07-15
feat(ObjectPage & DynamicPage): add data attributes to elements

Lukas742:


MarcusNotheis (approved):
many `data-component-name`s :)

PR #1848, opened 2021-07-20
chore(SF-0.131): downport consolidates API of ObjectPage and DynamicPage

Lukas742:
BREAKING CHANGE: DynamicPageHeader: children are no longer displayed as flex items to support other display types like grid. To align children you now need to add the container (like FlexBox) and CSS yourself.
BREAKING CHANGE: DynamicPageTitle: subHeading has been renamed. Please use subheading instead.
BREAKING CHANGE: DynamicPage: header has been renamed. Please use headerContent instead.
BREAKING CHANGE: DynamicPage: title has been renamed. Please use headerTitle instead.
BREAKING CHANGE: ObjectPage: title has been renamed to headerTitle and is now defining the upper, static, title section of the ObjectPage. It expects to receive the DynamicPageTitle component.
BREAKING CHANGE: ObjectPage: noHeader has been removed. It is now sufficient not to set headerTitle and header to achieve the same behavior.
BREAKING CHANGE: ObjectPage: title, subTitle, headerActions, breadcrumbs and keyInfos should now be passed to the corresponding DynamicPageTitle props.
BREAKING CHANGE: ObjectPageSection: title and titleUppercase has been renamed. Please use heading and headingUppercase instead.
BREAKING CHANGE: ObjectPageSubSection: title has been renamed. Please use heading instead.
BREAKING CHANGE: getScrollBarWidth has been removed.


packages/base/src/dist/Utils.ts
```diff
@@ -0,0 +1,3 @@
+import { deprecationNotice, enrichEventWithDetails } from '../utils';
```
MarcusNotheis, line 1: I think we don't have a dist folder for this release?

packages/main/src/components/DynamicPage/index.tsx
```diff
@@ -1,3 +1,4 @@
+import { enrichEventWithDetails } from '@ui5/webcomponents-react-base/dist/Utils';
```
MarcusNotheis, line 1: This should import from `lib`:
```suggestion
import { enrichEventWithDetails } from '@ui5/webcomponents-react-base/lib/Utils';
```

MarcusNotheis (changes requested):

packages/base/src/dist/Utils.ts
Lukas742: Done

packages/main/src/components/DynamicPage/index.tsx
Lukas742: Done

MarcusNotheis (approved):

PR #1857, opened 2021-07-22
fix(ObjectPage): make sure TabBar is always on top of content

### LINKED ISSUE #1856: ObjectPage: Tables overlap tab bar
**Describe the bug**
Putting a `Table` into an `ObjectPageSection` will make that table overlap the `ObjectPage` tab bar when scrolling.

**Isolated Example**
[Isolated example on `codesandbox.io`](https://codesandbox.io/s/prod-hill-4v223?file=/src/App.js).

**Expected behavior**
Table goes underneath header element instead of above it.

**Screenshots**
image

`@ui5/webcomponents` version: 1.0.0-rc14
`@ui5/webcomponents-react` version: 0.17.1
Operating System: macOS 11.4
Browser: Chrome 91

MarcusNotheis:
closes #1856

### DIFF
--- a/packages/main/src/components/ObjectPage/index.tsx
+++ b/packages/main/src/components/ObjectPage/index.tsx
@@ -705,9 +705,9 @@ const ObjectPage: FC<ObjectPagePropTypes> = forwardRef((props: ObjectPagePropTyp
         </div>
       )}
       <div
+        className={classes.anchorBar}
         data-component-name="ObjectPageTabContainer"
         style={{
-          position: 'sticky',
           top:
             headerPinned || scrolledHeaderExpanded
               ? `${topHeaderHeight + headerContentHeight}px`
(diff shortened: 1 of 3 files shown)


packages/main/src/components/ObjectPage/index.tsx
```diff
@@ -705,9 +705,9 @@ const ObjectPage: FC<ObjectPagePropTypes> = forwardRef((props: ObjectPagePropTyp
         </div>
       )}
       <div
+        className={classes.anchorBar}
         data-component-name="ObjectPageTabContainer"
         style={{
-          position: 'sticky',
```
Lukas742, line 710: We need `zIndex: 1` here, otherwise the lower part of the buttons is cut off. In my opinion it would make sense to create a new style class for this. WDYT
Screenshot 2021-07-22 at 22 56 37

Lukas742 (changes requested):

packages/main/src/components/ObjectPage/index.tsx
MarcusNotheis: you're absolutely right :) done

Lukas742 (approved):

Issue #1057, opened 2020-12-04
How to use `setTheme` in typescript?


auryn31:
**Describe the documentation issue**

- [x] Documentation is unclear
- [x] Documentation is missing

**Description of the issue**
I have issues to use `setTheme()` in typescript. 

```js
// next line is not importable
import { setTheme } from '@ui5/webcomponents-base/dist/config/Theme';

import '@ui5/webcomponents/dist/Assets';
import '@ui5/webcomponents-fiori/dist/Assets';
```

and i cannot find any setTheme in the ts files :-/

Maybe someone knows the solution?

Tanks for the help

MarcusNotheis:
Hey @auryn31,
I guess this is caused by the fact that the UI5 Web Components themselves do not offer any d.ts files as of now, only our "for React" wrapper has type definitions. As far as I know, the UI5 Web Components team is currently working on providing type definitions as well.
As a quick fix for know: Is it possible to make your tsconfig less strict? You can find our [tsconfig here](https://github.com/SAP/ui5-webcomponents-react/blob/master/tsconfig.json) and we can import `useTheme` without issues.

auryn31:
@MarcusNotheis  Thanks for the advice :-)
I will test this. 
If the team needs help with the types, i think we also can help if there is time :-)
Where can i find the repo for the types? Then i can create a pr if i found it :-)

MarcusNotheis:
I think the d.ts files will be generated from the jsdoc comments, but I guess @ilhan007 can explain their plan in a more detailed way :)

MarcusNotheis:
Closing this issue in favor of #1136.
Please feel free to continue the conversation there.

PR #1977, opened 2021-08-12
chore(base): create type definitions for theme configuration

Psidium:
Hey, following what I've seen in #1057 and #1136, I believe we could follow the same approach we did for  `UI5Devices.d.ts` and create a type definition for the base Theming config. 

I'm using these types in my project (internally at SAP).

I created the themes as an interface only so the user can use https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation to add a new theme in their project without disrupting the already established themes, like so:

```ts
declare module "@ui5/webcomponents-base/dist/config/Theme" {
  export interface ThemeKeys {
    my_new_theme: never;
  }
}
```

The only issue is with the themes themselves, since that when a new theme is created this file here will have to be updated.

### DIFF
--- a/packages/base/types/Theme.d.ts
+++ b/packages/base/types/Theme.d.ts
@@ -8,10 +8,14 @@ declare module '@ui5/webcomponents-base/dist/config/Theme' {
     | 'sap_fiori_3_hcb'
     | 'sap_fiori_3_hcw'
     | string;
-
+  
+  /**
+   * Getter for the currenty applied theme.
+   */ 
   export function getTheme(): ThemeId;
 
+  /**
+   * Apply a new theme to the application.
+   */ 
   export function setTheme(theme: ThemeId): Promise<void>;
-
-
 }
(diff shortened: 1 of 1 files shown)


Psidium:
Another option would be to 

```ts
import { Themes } from "../../main/src/dist/Themes";

declare module "@ui5/webcomponents-base/dist/config/Theme" {
    export function getTheme(): Themes;
    export function setTheme(theme: Themes): void;
}
```

MarcusNotheis:
Hey @Psidium,
thanks for creating this PR! I tried to take this one step further and integrate the type definitions directly in the UI5 Web Components Repo: https://github.com/SAP/ui5-webcomponents/pull/3653
Let's see if the Web Components PR will be accepted, otherwise we can go on and merge this one 🙂

Psidium:
Regarding the `getTheme` function... shouldn't I use `useTheme` from the container used in `ThemeProvider` so that my component is aware of theme changes?

MarcusNotheis:
Thanks for kicking this off, @Psidium! I just added `string` as allowed value for the setTheme function as well so that custom themes via SAP Theme Designer are supported as well.

MarcusNotheis (approved):

PR #2100, opened 2021-09-08
fix(Bar & ColumnChart): always show labels if enough space is available

Lukas742:
When quickly changing the dimensions of the Bar or ColumnChart, sometimes the labels weren't displayed. This is because of the `Animate` component of `react-smooth` not calling the `onAnimationEnd` callback in some cases. There is already an open issue, but it's pretty old: https://github.com/recharts/recharts/issues/2132
I tried to reproduce the issue with plain `recharts` components, but wasn't able to, that's why I couldn't create a PR in their repo.

This PR adds a workaround, so that the bars always receive an update when they are animated. If bars don't call the `onAnimationEnd` callback after 1.5 seconds, a rerender is forced and therefore updating the bars and the labels within.


packages/charts/src/hooks/useCancelAnimationFallback.ts
```diff
@@ -0,0 +1,23 @@
+import { debounce } from '@ui5/webcomponents-react-base/dist/Utils';
+import { useRef, useState } from 'react';
+
+export const useCancelAnimationFallback = (noAnimation) => {
```
MarcusNotheis, line 4:
```suggestion
export const useCancelAnimationFallback = (noAnimation: boolean) => {
```
Maybe we can add some strict types here?

MarcusNotheis (approved):

packages/charts/src/hooks/useCancelAnimationFallback.ts
Lukas742: Done, also added the return type.

PR #2275, opened 2021-11-03
feat: add `portalContainer` prop for elements rendered via `createPortal`

Lukas742:
This PR adds the `portalContainer` prop to components wich are using `React.createPortal` to render modals or other elements outside of the DOM hierarchy. With `portalContainer` it is possible to change the element the modal is rendered into.

### DIFF
--- a/packages/main/src/components/FilterBar/FilterDialog.tsx
+++ b/packages/main/src/components/FilterBar/FilterDialog.tsx
@@ -255,7 +255,7 @@ export const FilterDialog = (props) => {
         );
       });
   };
-  console.log(portalContainer);
+
   return createPortal(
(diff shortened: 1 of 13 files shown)


packages/main/src/components/FilterBar/FilterDialog.tsx
```diff
@@ -254,7 +255,7 @@ export const FilterDialog = (props) => {
         );
       });
   };
-
+  console.log(portalContainer);
```
MarcusNotheis, line 258: please remove console.log

MarcusNotheis (changes requested):

MarcusNotheis (approved):

PR #2314, opened 2021-11-15
fix: sync refs correctly with `useSyncRef` hook

Lukas742:
This PR fixes the issue of `useConsolidatedRef` returning a ref which is one cicle "late". It implements the new `useSyncRef` hook, which is now returning the actual current node. `useConsolidatedRef` is therefore now deprecated and will be removed with the next minor release.


packages/base/src/hooks/useSyncRef.ts
```diff
@@ -0,0 +1,19 @@
+import { MutableRefObject, Ref, RefCallback, RefObject, useRef } from 'react';
+
+export function useSyncRef<RefType = any>(ref: Ref<RefType>): [RefCallback<RefType>, RefObject<RefType>] {
+  const localRef = useRef<RefType>();
```
MarcusNotheis, line 4:
```suggestion
  const localRef = useRef<RefType>(null);
```
I guess this should be called like this to return a `MutableRefObject`

packages/base/src/hooks/useSyncRef.ts
```diff
@@ -0,0 +1,19 @@
+import { MutableRefObject, Ref, RefCallback, RefObject, useRef } from 'react';
+
+export function useSyncRef<RefType = any>(ref: Ref<RefType>): [RefCallback<RefType>, RefObject<RefType>] {
```
MarcusNotheis, line 3:
```suggestion
export function useSyncRef<RefType = any>(ref: Ref<RefType>): [RefCallback<RefType>, MutableRefObject <RefType>] {
```
isn't this rather a MutableRefObject?

packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -539,7 +540,10 @@ const AnalyticalTable = forwardRef((props: AnalyticalTablePropTypes, ref: Ref<HT
     return props.data;
   }, [props.data, minRows]);
 
-  const tableInstanceRef = useConsolidatedRef<Record<string, any>>(tableInstance);
+  let tableInstanceRef = useRef<Record<string, any>>();
+  if (tableInstance) {
+    tableInstanceRef = tableInstance;
+  }
```
MarcusNotheis, line 546: This is technically a breaking change - we have accepted a callback ref for the `tableInstance` before that change

packages/main/src/components/FilterBar/FilterBar.test.tsx
```diff
@@ -500,7 +500,7 @@ describe('FilterBar', () => {
     const dialogGoButton = screen.getAllByText('Go')[1];
     fireEvent.click(dialogGoButton);
     expect(onGo).toHaveBeenCalledTimes(2);
-    expect(onFiltersDialogSave).toHaveBeenCalledTimes(1);
+    expect(onFiltersDialogSave).toHaveBeenCalledTimes(2);
```
MarcusNotheis, line 503: why did those counters increase?

Lukas742: Before, onGo just closed the dialog and din't update the values inside the FilterBar. Now, when the "Go" button is clicked, it's also saving the values and therefore is also calling the corresponding callback.

packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -539,7 +540,7 @@ const AnalyticalTable = forwardRef((props: AnalyticalTablePropTypes, ref: Ref<HT
     return props.data;
   }, [props.data, minRows]);
 
-  const tableInstanceRef = useConsolidatedRef<Record<string, any>>(tableInstance);
+  let tableInstanceRef = useRef<Record<string, any>>();
```
MarcusNotheis, line 543:
```suggestion
  const tableInstanceRef = useRef<Record<string, any>>(null);
```
`tableInstanceRef` is never reassigned, so I would prefer const. In addition, using null as default value makes it a `MutableRefObject` :)

PR #2320, opened 2021-11-16
refactor(AnalyticalTable): deprecate column option `canReorder` in favor of `disableDragAndDrop`

### LINKED ISSUE #2318: [SF] [AnalyticalTable] Is Analytical Table disabled to change column order by Drag and Drop?
**Describe the bug**
Users can drag the column header and drop to change the column's order. It's a good feature but it's unexpected for us. Does the feature be disabled by setting one property? I didn't find it in the document. 

Thank you,
Christy

Lukas742:
This PR deprecates the `canReorder` column option in favor of the more expressive `disableDragAndDrop` option. Also it removes the column option `canResize` from the public documentation as it's only used internally by the react-table and had no effect when set as column option. To disable columns from resizing use `disableResizing` instead.

Closes #2318


packages/main/src/components/AnalyticalTable/AnalyticalTable.stories.mdx
```diff
@@ -188,6 +188,8 @@ import { useRowDisableSelection } from '@ui5/webcomponents-react/dist/Analytical
...
+| `disableDragAndDrop`  | `boolean`                                                                | Defines if the column is reorderable by dragging and dropping columns.                                                                                                                                                                                                                                                                                                                                          |
+| `canReorder`          | `boolean`                                                                | **deprecated** Defines if the column is reorderable by dragging and dropping columns. Defaults to: `true`.                                                                                                                                                                                                                                                                                                      |
```
MarcusNotheis, line 192:
```suggestion
| `canReorder`          | `boolean`                                                                | **deprecated** Defines if the column is reorderable by dragging and dropping columns. Please use `disableDragAndDrop` instead. Defaults to: `true`.                                                                                                                                                                                                                                                                                                      |
```

MarcusNotheis (changes requested):

MarcusNotheis (approved):

PR #2383, opened 2021-12-07
chore: update to ui5-wc 0.31.27 & downport changes

Lukas742:


packages/main/src/components/DynamicPage/index.tsx
```diff
@@ -7,6 +7,7 @@ import { FlexBox } from '@ui5/webcomponents-react/lib/FlexBox';
 import { GlobalStyleClasses } from '@ui5/webcomponents-react/lib/GlobalStyleClasses';
 import { PageBackgroundDesign } from '@ui5/webcomponents-react/lib/PageBackgroundDesign';
 import { CommonProps } from '@ui5/webcomponents-react/interfaces/CommonProps';
+import debounce from 'lodash/debounce';
```
MarcusNotheis, line 10: we should rather use our debounce?

MarcusNotheis (changes requested):

MarcusNotheis (approved):

packages/main/src/components/DynamicPage/index.tsx
MarcusNotheis: nevermind 🙂

PR #2480, opened 2022-01-10
fix: support `title` attribute by all components

### LINKED ISSUE #2472: [Button] Native title attribute does not work.
**Describe the bug**
Hi team,
It seems you cannot use the native title attribute when using the react wrappers. The wrappers favor `tooltip` prop. The ui5-webcomponents accepts the `title` attribute, so when we switched to use the react wrappers, this causes a major breaking change for accessibility. Please see:  [CommonProps.ts#L17](https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/main/src/interfaces/CommonProps.ts#L17) and [withWebComponent.tsx#L127](https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/main/src/internal/withWebComponent.tsx#L127)
Thanks,
Allen

**Isolated Example**
[CodeSandbox](https://codesandbox.io/s/amazing-wilbur-myvs9?file=/src/App.js)

**To Reproduce**
1. Go to [CodeSandbox](https://codesandbox.io/s/amazing-wilbur-myvs9?file=/src/App.js)
2. Hover over each button and notice tooltip only displays for one of the buttons using the `tooltip` prop.
2. Inspect the dom and you will see the button with title attribute does not have the attribute set.

**Expected behavior**
Native `title` attribute should still work for react wrapper components. 

`@ui5/webcomponents` version: 1.0.x
`@ui5/webcomponents-react` version: 0.20.x

Lukas742:
This PR adds the support of the `title` attribute to all ui5-webcomponents. Also the `tooltip` prop is now deprecated and will be removed with the next minor version.

Fixes #2472

### DIFF
--- a/packages/main/src/interfaces/CommonProps.ts
+++ b/packages/main/src/interfaces/CommonProps.ts
@@ -12,7 +12,9 @@ export interface CommonProps extends HTMLAttributes<HTMLElement> {
    */
   className?: string;
   /**
-   * A tooltip which will be shown on hover
+   * A tooltip which will be shown on hover.
+   *
+   * @deprecated please use `title` attribute instead.
    */
   tooltip?: string;
   ref?: Ref<any>;
--- a/packages/main/src/interfaces/CommonProps.ts
+++ b/packages/main/src/interfaces/CommonProps.ts
@@ -13,8 +13,9 @@ export interface CommonProps extends HTMLAttributes<HTMLElement> {
   className?: string;
   /**
    * A tooltip which will be shown on hover.
+   * This prop is deprecated as of v0.20.6, please use the native `title` prop instead.
    *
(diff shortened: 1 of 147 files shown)


packages/main/src/interfaces/CommonProps.ts
```diff
@@ -12,7 +12,9 @@ export interface CommonProps extends HTMLAttributes<HTMLElement> {
    */
   className?: string;
   /**
-   * A tooltip which will be shown on hover
+   * A tooltip which will be shown on hover.
```
MarcusNotheis, line 15:
```suggestion
   * A tooltip which will be shown on hover.
   * *This prop is deprecated as of v0.20.6, please use the native `title` prop instead.*
```

MarcusNotheis (approved):

PR #2487, opened 2022-01-11
feat(SplitterLayout): initial component implementation

Lukas742:


packages/main/src/components/SplitterLayout/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * layout will be arranged in that way. In order to split vertically and horizontally at the same time, splitters need
+ * to be nested.
+ * By adding or changing `SplitterElement`s to the `SplitterLayout` that make up the content areas, the size can be changed
+ * programmatically. Additionally the content areas can be made non-resizable individually and a minimal size (in px)
+ * can be set.
+ * The splitter bars are focusable to enable resizing of the content areas via keyboard. The size of the content areas
+ * can be manipulated when the splitter bar is focused and Left/Down/Right/Up are pressed.
+ */
+const SplitterLayout = forwardRef((props: SplitterLayoutPropTypes, ref: Ref<HTMLDivElement>) => {
+  const { vertical = false, children, title, tooltip, style, className, ...rest } = props;
```
MarcusNotheis, line 38: We usually used `SplitterLayout.defaultsProps = {}`. Any specific reason to use the ES6 default here?

packages/main/src/components/SplitterLayout/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const classes = useStyles();
+
+  const layoutElements = useConcatSplitterElements(children ?? [], style?.width, style?.height, vertical);
```
MarcusNotheis, line 41: Optional:
Should we refactor this function to accept on object instead of four parameters? e.g:
```suggestion
  const layoutElements = useConcatSplitterElements({ children: children ?? [], width: style?.width, height: style?.height, vertical });
```
This would increase readability IMO

packages/main/src/dist/SplitterElement.ts
```diff
@@ -0,0 +1,5 @@
+import { SplitterElement } from '../components/SplitterElement';
+import { SplitterElementPropTypes } from '../components/SplitterElement';
```
MarcusNotheis, line 2:
```suggestion
import type { SplitterElementPropTypes } from '../components/SplitterElement';
```

scripts/rollup/configFactory.js
```diff
@@ -34,6 +34,7 @@ const rollupConfigFactory = (pkgName, externals = []) => {
       'react',
       'react-dom',
       'react-jss',
+      'prop-types',
```
MarcusNotheis, line 37: Why do we need to add prop-types here?

packages/main/src/internal/safeGetChildrenArray.ts
```diff
@@ -0,0 +1,4 @@
+import { Children, ReactNode } from 'react';
+
+export const safeGetChildrenArray = <T = Exclude<ReactNode, boolean | null | undefined>>(children: unknown): T[] =>
```
MarcusNotheis, line 3: Should we change this implementiation to:
```
  // Not using React.Children.toArray because it mutates the key prop.
  let childArray: ReactElement[] = [];
  React.Children.forEach(children, child => {
    if (React.isValidElement(child)) {
      childArray.push(child);
    }
  });
```
If yes, then we should also credit react-spectrum for that: https://github.com/adobe/react-spectrum/blob/dfe72ee076f7ca27728d0b6b4c1ef86e11191ece/packages/%40react-spectrum/breadcrumbs/src/Breadcrumbs.tsx#L41-L47
and add it to the reuse file

packages/main/src/components/SplitterLayout/SplitterLayout.stories.mdx
```diff
@@ -0,0 +1,157 @@
...
+          <SplitterElement resizable={false}>
+            <Button style={{ width: '100%' }}>Content 4 (not resizable)</Button>
+          </SplitterElement>
+          <SplitterElement>
+            <Button style={{ width: '100%' }}>Content 5</Button>
+          </SplitterElement>
+          <SplitterElement>
+            <Button style={{ width: '100%' }}>Content 6 (size: 200px)</Button>
+          </SplitterElement>
+        </SplitterLayout>
```
MarcusNotheis, line 51: Although this works as intended I would try to keep it a bit simpler. I tried to move around elements and directly hit the min-size or the non-resizable element which feels buggy for the end user.
Maybe only add one element that has a min Size and a max size, the other elements can resize freely?
And add a second story for non-resizable elements?

packages/main/src/components/SplitterLayout/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useConcatSplitterElements } from './useConcatSplitterElements';
+import clsx from 'clsx';
+
+const useStyles = createUseStyles(styles, { name: 'SplitterLayout' });
+
+export interface SplitterLayoutPropTypes extends CommonProps {
+  /**
+   * Controls if a vertical or horizontal `SplitterLayout` is rendered.
+   */
+  vertical?: boolean;
```
MarcusNotheis, line 18: This prop behaves incorrect in my opinion:
image

This should be `vertical=true`, not `false`.

MarcusNotheis (changes requested):

packages/main/src/components/SplitterLayout/index.tsx
Lukas742: Imo this shouldn't be set at all, since `undefined` is not necessarily the same as `false`.

packages/main/src/components/SplitterLayout/index.tsx
Lukas742: Yes, you are right. I compared our implementation with the one of [SAPUI5](https://sapui5.hana.ondemand.com/#/entity/sap.ui.layout.Splitter/sample/sap.ui.layout.sample.Splitter) and there it's implemented the other way around. (The default for `orientation` is `"Horizontal"` so the splitter elements in the demo are initially horizontally aligned)

packages/main/src/components/SplitterLayout/SplitterLayout.stories.mdx
Lukas742: Done.

MarcusNotheis (approved):
LGTM 🚢

PR #2525, opened 2022-01-24
refactor(RadialChart): correct `chartConfig` types

Lukas742:


MarcusNotheis (approved):
as commit message: maybe rather `fix(RadialChart)`?

PR #2654, opened 2022-03-04
feat: add `Modals` component for opening modals imperatively

### LINKED ISSUE #1987: Proposal: Imperative methods of MessageBox, Dialog, Toast

No.

It's tedious to use the above components in the scenario that multiple (MessageBox/Dialog/Toast) components in the same page. Users may need to use multiple intermediate variables to manifest the show/off status.

Imagine a situation: we need to display a MessageBox after an API response, we may need to store at least two variables: `MessageBoxOpen` & `apiResponse.rejectedReason`. Both of them are irrelevant to the main logic and perplex our codebase.

```tsx
<MessageBox open={messageOneOpened}>{apiResponse.rejectedReason}</MessageBox>
<MessageBox open={messageTwoOpened}>{apiResponse.rejectedReason}</MessageBox>
<MessageBox open={messageThreeOpened}>{apiResponse.rejectedReason}</MessageBox>
```

Learn from [other UI libraries](https://ant.design/components/modal/), we can use the imperative method to simplify the procedure.

```tsx
Modal.info({
  title: 'This is a notification message',
  content: (
    <div>
      <p>some messages...some messages...</p>
      <p>some messages...some messages...</p>
    </div>
  ),
  onOk() {},
});
```

<details><summary>CLICK ME</summary>
<p>

```tsx
import React from 'react';
import ReactDOM from 'react-dom';

const CONTAINER_ID = 'message-box-container';

let mfeContainer: HTMLElement | null = null;

export function createMessageBoxContainer(container: HTMLElement) {
  const div = document.createElement('div');
  mfeContainer = container;
  div.id = CONTAINER_ID;
  container.appendChild(div);
}

export function removeMessageBoxContainer() {
  const div = mfeContainer?.querySelector<HTMLElement>(`#${CONTAINER_ID}`) as HTMLElement;
  ReactDOM.unmountComponentAtNode(div);
}

/**
 * @example
 * messageBox(
 *   <MessageBox
 *     actions={['Cancel']}
 *     onClose={() => {
 *       alert('onClose');
 *     }}
 *   >
 *     MessageBox
 *   </MessageBox>,
 * );
 */
export default function messageBox(element: React.ReactElement) {
  const clonedElement = React.cloneElement(element, {
    open: true,
    onClose: () => {
      element.props?.onClose?.();
      removeMessageBoxContainer();
    },
  });
  ReactDOM.render(clonedElement, mfeContainer?.querySelector(`#${CONTAINER_ID}`) as HTMLElement);
}
```

</p>
</details>

**Additional context**
None.

MarcusNotheis:
To Do:

- [x] Document Return Types
- [x] Document function signature
- [x] Tests

Closes #1987

### DIFF
--- a/packages/main/src/components/Modals/index.tsx
+++ b/packages/main/src/components/Modals/index.tsx
@@ -1,55 +1,173 @@
-import React, { createRef } from 'react';
+import React, { createRef, ElementType, RefObject } from 'react';
 import ReactDOM from 'react-dom';
 import {
   Dialog,
+  DialogDomRef,
   DialogPropTypes,
+  Popover,
+  PopoverDomRef,
   PopoverPropTypes,
+  ResponsivePopover,
+  ResponsivePopoverDomRef,
   ResponsivePopoverPropTypes,
   Toast,
   ToastDomRef,
   ToastPropTypes
 } from '../../webComponents';
-import { MessageBoxPropTypes } from '../MessageBox';
+import { MessageBox, MessageBoxPropTypes } from '../MessageBox';
+import { ThemeProvider } from '../ThemeProvider';
+
+type ModalReturnType<DomRef> = {
+  ref: RefObject<DomRef>;
+};
+
+type ClosableModalReturnType<DomRef> = ModalReturnType<DomRef> & {
+  close: () => void;
+};
+
+const getContainer = (givenContainer: HTMLElement) => {
+  if (givenContainer) {
+    return givenContainer;
+  }
+  const container = document.createElement('div');
+  document.body.appendChild(container);
+  return container;
+};
 
 const unmountComponent = (container: DocumentFragment | HTMLElement) => {
   ReactDOM.unmountComponentAtNode(container);
   container.parentElement.removeChild(container);
 };
 
-export function showDialog(props: DialogPropTypes) {
-  setTimeout(() => {
-    const container = document.createDocumentFragment();
-    ReactDOM.render(
-      <Dialog
-        {...props}
-        open
-        onAfterClose={(event) => {
-          if (typeof props.onAfterClose === 'function') {
-            props.onAfterClose(event);
-          }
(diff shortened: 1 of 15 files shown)


packages/main/src/components/Modals/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+const popupOnAfterCloseFactory = (props, container) => (event) => {
+  if (typeof props.onAfterClose === 'function') {
+    props.onAfterClose(event);
+  }
+  unmountComponent(container);
+};
+
+const render = (Element: ElementType, props: any, container: HTMLElement): Promise<void> => {
+  return new Promise((resolve) => {
```
mauroerta, line 53: Can I ask why is needed to create a `Promise` and render the element inside a `setTimeout`? 
Also, I wanted to ask you if could be an idea to use a [Portal](https://reactjs.org/docs/portals.html) in this circumstance.

MarcusNotheis: Good catch with the Promise - this was some earlier experiment which is removed in the meantime. So I've removed the promise implementation as well.
I'm using `setTimeout` here because ReactDOM.render seems to be blocking other react events when called synchronously, so I decided to make this asynchronous.

We can't use `createPortal` here because it doesn't really render the element into the DOM but is rather an escape hatch from the current DOM structure while rendering.

mauroerta: Makes sense, thank for the clarifications.

MarcusNotheis: You're welcome 🙂 
And I have to correct myself: the promise is needed here:
https://github.com/SAP/ui5-webcomponents-react/blob/bdc42404876827dfd7bdaea1510bf68e42460501/packages/main/src/components/Modals/index.tsx#L175-L180

Lukas742 (approved):

PR #2660, opened 2022-03-08
fix(Toolbar): spacings between elements

### LINKED ISSUE #2032: [ObjectPage]: Header action bar is not responsive when overflow button exist
Hi Colleague,
In the header of Object page component, when there are common button and overflow button, the layout is not responsive, see more details in following screenshots. 

### **Isolated Example**
https://codesandbox.io/embed/nifty-pine-rtx97?file=/src/App.js&codemirror=1

**Screenshots**

image

image

`@ui5/webcomponents` version: 0.31.15
`@ui5/webcomponents-react` version:0.131.15

**Expected behavior**
image

**Additional context**
Could you take a look and we hope that it will be fixed in next release 0.131.x. 
BTW, changing "visibility:hidden" to "display: none" is work as expected in my case.

mauroerta:
## Proposed change

This is the easier and shorter way I found to fix the bug explained [here](https://github.com/SAP/ui5-webcomponents-react/issues/2032).

```diff
- style: { visibility: 'hidden' }
+ style: { visibility: 'hidden', position: 'absolute', pointerEvents: 'none' }
```

## The problem

The elements inside the toolbar can be hidden based on the size of the window, when they're hidden the CSS property `visibility` is set to `hidden` - But even if they are NOT visible though, they still occupy space, and this is the cause of the bug.
In my solution, other than `visibility: hidden` I also set `position: absolute`, this way we can be sure that the hidden elements don't occupy space, `pointer-events: none` ensures that they will not be clickable in any case.

## Alternatives I've considered

Instead of `visibility: hidden` I tried using `display: none`, but this appears to doesn't work as expected, in fact, it causes an infinite re-render where we can see the item visible for a few milliseconds, then hidden for another few milliseconds and again visible... This issue is probably caused by the `ResizeObserver`.
Force the width to 0 causes the same issue.

## Further comments

I think the proposed solution would be just a temporary solution, Probably, it's better to think about a complete refactor of the Toolbar component.

Closes #2032

## Pull Request Checklist
- [x] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents-react/blob/master/CONTRIBUTING.md)
- [x] [Correct commit message style](https://github.com/SAP/ui5-webcomponents-react/blob/master/docs/Guidelines.md#commit-message-style)

### DIFF
--- a/packages/main/src/components/Toolbar/__snapshots__/Toolbar.test.tsx.snap
+++ b/packages/main/src/components/Toolbar/__snapshots__/Toolbar.test.tsx.snap
@@ -251,7 +251,7 @@ exports[`Toolbar overflow menu 1`] = `
         <div
           class="Toolbar-childContainer"
           data-component-name="ToolbarChildContainer"
-          style="visibility: hidden;"
+          style="visibility: hidden; position: absolute; pointer-events: none;"
         >
           <span
             class="Text-text"
@@ -264,7 +264,7 @@ exports[`Toolbar overflow menu 1`] = `
         <div
           class="Toolbar-childContainer"
           data-component-name="ToolbarChildContainer"
-          style="visibility: hidden;"
+          style="visibility: hidden; position: absolute; pointer-events: none;"
         >
           <span
             class="Text-text"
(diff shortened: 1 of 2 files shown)


Lukas742 (approved):
LGTM

As you already mentioned we cannot use `display: none;` as we need to know the position of the last visible element which we currently get with `lastElement.getBoundingClientRect().right` in the observer. 

A refactoring is probably inevitable if we want to offer all UX features the Fiori Design guidelines describe. Right now we just don't have the capacity to work on this.

PR #2701, opened 2022-03-23
fix(VariantManagement): call onSaveManageViews only with callbackProperties

### LINKED ISSUE #2714: VariantManagement: Calling callback too early
**Describe the bug**
Callback is being called and then parameter of the callback function is updated/mutates.

**Expected behavior**
Only call the callback once its parameter has its properties set and do not mutate it after the call.

`@ui5/webcomponents` version: 1.2.1
`@ui5/webcomponents-react` version: 0.22.2
Operating System: MAC
Browser: Chrome

**Additional context**
[Pull request](https://github.com/SAP/ui5-webcomponents-react/pull/2701)

guedes4g:
This Pr fixes VariantManagement onSaveManageViews being called with empty callback properties

Fixes #2714


packages/main/src/components/VariantManagement/index.tsx
```diff
@@ -353,6 +350,9 @@ const VariantManagement = forwardRef((props: VariantManagementPropTypes, ref: Re
             });
           }
           callbackProperties.variants.push({ ...child.props, ...updatedProps, variantItem: currentVariant });
+          if (typeof onSaveManageViews === 'function') {
+            onSaveManageViews(enrichEventWithDetails(e, callbackProperties));
+          }
```
Lukas742, line 355: Please call the function outside of the `map` and setter.

Lukas742 (changes requested):

packages/main/src/components/VariantManagement/index.tsx
guedes4g: Moved

Lukas742 (approved):
LGTM! Thanks for your contribution :)

PR #2713, opened 2022-03-28
feat(SplitterLayout): add `options` prop to customize reset behavior, correct sizing of elements

Lukas742:
This PR adds the possibility to control whether the `SplitterLayout` should be reset when dependencies, children or sizing changes. It also fixes an issue with conditional rendering and sizing.


packages/main/src/components/SplitterLayout/index.tsx
```diff
@@ -45,8 +79,51 @@ const SplitterLayout = forwardRef((props: SplitterLayoutPropTypes, ref: Ref<HTML
     vertical
   });
 
+  useEffect(() => {
+    if (!initialChildren.current && options?.resetOnChildrenChange) {
+      setReset(true);
+    }
+    initialChildren.current = false;
+  }, [children]);
```
MarcusNotheis, line 87:
```suggestion
  }, [children, options?.resetOnChildrenChange]);
```

packages/main/src/components/SplitterLayout/index.tsx
```diff
@@ -45,8 +79,51 @@ const SplitterLayout = forwardRef((props: SplitterLayoutPropTypes, ref: Ref<HTML
...
+
+  useEffect(() => {
+    if (!initialCustomDep.current && options?.resetOnCustomDepsChange) {
+      setReset(true);
+    }
+    initialCustomDep.current = false;
+  }, options?.resetOnCustomDepsChange ?? []);
```
MarcusNotheis, line 94:
```suggestion
  }, [options?.resetOnCustomDepsChange, ...(options?.resetOnCustomDepsChange ?? []));
```

packages/main/src/components/SplitterElement/index.tsx
```diff
@@ -51,20 +51,49 @@ export interface SplitterElementPropTypes extends CommonProps {
  */
 const SplitterElement = forwardRef((props: SplitterElementPropTypes, ref: RefObject<HTMLDivElement>) => {
   const { children, style, tooltip, className, minSize, size, resizable, ...rest } = props;
+  const [componentRef, splitterElementRef] = useSyncRef(ref);
+  const { vertical, reset } = useContext(SplitterLayoutContext);
+  const [flex, setFlex] = useState(size !== 'auto' ? `0 0 ${size}` : '1 0 min-content');
```
MarcusNotheis, line 56: `min-content` doesn't work for Safari: https://caniuse.com/mdn-css_properties_flex-basis_min-content

packages/main/src/components/SplitterElement/index.tsx
```diff
@@ -51,20 +51,49 @@ export interface SplitterElementPropTypes extends CommonProps {
...
+  const [flexBasisApplied, setFlexBasisApplied] = useState(false);
```
MarcusNotheis, line 57: I'm not 100% why we need the second state here - can't we derive it from size and reset?

MarcusNotheis (changes requested):

MarcusNotheis (approved):

PR #2796, opened 2022-04-15
feat: update @ui5/webcomponents to 1.3.0

MarcusNotheis:
- [x] Update Peer Dependencies
- [x] Update Templates
- [x] Update to final @ui5/webcomponents 1.3.0

Required Pull Requests
- https://github.com/SAP/ui5-webcomponents/pull/5064

BREAKING CHANGE: __FilterItem__: The FilterItem now extends the UI5Element and not the ListItem anymore. As a result the `type`, `selected` and `onDetailClick` props have been removed.


MarcusNotheis:
@ilhan007 looks like there are some breaking changes included in `ui5-filter-item` component. Is this intended?
image

ilhan007:
> @ilhan007 looks like there are some breaking changes included in `ui5-filter-item` component. Is this intended? image

Yes @MarcusNotheis it is intended, it's not done by mistake. The FilterItem now extends the UI5Element, used to extend the ListItem as it turns out that most of the code inherited from the ListItem are not used by the FilterItem and we did this refactoring, which is technically breaking change. However, on our side the FIlterItem API was never documented, the samples are using just the "text" property, which will continue to work, so in case it does not do much harm on your side, we would like to keep it.

Lukas742 (approved):

PR #2909, opened 2022-05-03
Chore/remove deprecated

alina-hs:
Remove deprecated: 
- tooltip
- useConsolidatedRef 
- StyleClassHelper


packages/main/src/components/MessageBox/index.tsx
```diff
@@ -235,7 +231,7 @@ const MessageBox = forwardRef((props: MessageBoxPropTypes, ref: Ref<DialogDomRef
       slot={slot}
       ref={ref}
       style={style}
-      title={tooltip ?? props.title}
+      title={props.title}
```
Lukas742, line 234: We don't need to set `title` here anymore, please remove it here and in the destructed assignment (`  const { footer, headerText, title, onAfterClose, ...restWithoutOmitted } = rest;`)

packages/main/src/components/ResponsiveGridLayout/index.tsx
```diff
@@ -87,20 +86,17 @@ const ResponsiveGridLayout = forwardRef((props: ResponsiveGridLayoutPropTypes, r
...
   const finalClassNames = clsx(classes.container, className);
 
-  useDeprecationNoticeForTooltip('ResponsiveGridLayout', props.tooltip);
-
   return (
     <div
       ref={ref}
       className={finalClassNames}
-      title={tooltip ?? title}
+      title={title}
```
Lukas742, line 99: `title` is no longer necessary here, please remove it here and also from props.

Lukas742 (changes requested):
Please add an entry to our [Migration Guide](https://github.com/SAP/ui5-webcomponents-react/blob/main/docs/2-MigrationGuide.stories.mdx) stating what has been removed and what to use instead. 

Please also remove the tooltip from the wrapper script:
https://github.com/SAP/ui5-webcomponents-react/blob/da43383f6a4402636b671c2d296710d483150b5c/packages/main/scripts/create-web-components-wrapper.mjs#L381

PR #2927, opened 2022-05-09
fix(VariantManagement): correct `valueState` behavior

### LINKED ISSUE #2877: VariantManagement: "A file with this name already exists" error
**Describe the bug**
Currently, when using the 'VariantManagement' component, whenever you try to save a new variant using the "saveAs" function it raises an error saying "A file with this name already exists" even if you don't have created a variant with that name. It is occurring at the first time you click the 'saveAs' button, but if you click it again then it is saved. Also if the user focus out (click outside the input field) before saving, the error is not displayed and the new variant is normally saved.

**To Reproduce**
1. Render the 'VariantManager' component with a default 'VariantItem'
2. Click on 'saveAs' button
3. Write a new name for the variant
4. Click on the save button, you will receive the error
5. If you click it again, then it will save normally

**Expected behavior**
When clicking at the save button, after clicking on the 'saveAs' button, it is expected that a variant is saved without errors. Of course assuming that the name is new, in other words, that there is no variant with the same name listed

**Screenshots**
Here a screenshot with the created variants under "My Views" and the modal trying to create a new variant
VariantManagement 1

`@ui5/webcomponents` version: "1.2.3"
`@ui5/webcomponents-react` version: "0.22.4"
Operating System: Macos
Browser: Chrome

alina-hs:
This PR fixes some problems in the VariantManagement component.

In ManageViewsDialog
-	Change event 'onChange' to 'onInput' in Input component (Fixes #2877)
-	Remove padding from Footer in ManageViewsDialog for better design 

In SaveViewDialog
-	An invalid variantName is now marked as invalid directly on input and not only on save.


packages/main/src/components/VariantManagement/ManageViewsDialog.tsx
```diff
@@ -32,6 +32,14 @@ addCustomCSSWithScoping(
...
   `
 );
+addCustomCSSWithScoping(
+  'ui5-dialog',
+  `
+  :host([data-component-name="VariantManagementManageViewsDialog"]) .ui5-popup-footer-root{
+    padding: 0;
+  }
+  `
+);
```
Lukas742, line 42: Please use the [::part pseudo element](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) instead for both injected styles. 
--> remove `addCustomCSSWithScoping` for the content and the footer and add the styles with `createUseStyles` 
E.g.: 
```
   <className of Dialog>: {
      '&::part(content)': {
        padding: 0
      },
      '&::part(footer)': {
        padding: 0
      }
    },
```

packages/main/src/components/VariantManagement/ManageViewsDialog.tsx
```diff
@@ -159,7 +167,7 @@ export const ManageViewsDialog = (props: ManageViewsDialogPropTypes) => {
         deletedRows: deletedTableRows.current
       });
     } else {
-      Object.values(invalidVariants)[0].focus();
+      Object.values(invalidVariants).forEach((invalidVariant) => invalidVariant.focus());
```
Lukas742, line 170: Why has this been changed? This will set the focus on every invalid variant, which is not the desired behavior, since only one element should be focused at a time.

Lukas742 (changes requested):

packages/main/src/components/VariantManagement/ManageViewsDialog.tsx
alina-hs: The reason was because it seemed easier to see all invalid variants at the same time. Would you prefer the one at a time behaviour?

Lukas742: Yup, because there can only ever be one activeElement and so I think we should start with the first invalid variant.

packages/main/src/components/VariantManagement/ManageViewsDialog.tsx
```diff
@@ -14,24 +14,28 @@ import {
...
+import { ManageViewsTableRows } from './ManageViewsTableRows';
 import { VariantItemPropTypes } from './VariantItem';
 
-addCustomCSSWithScoping(
-  'ui5-dialog',
-  `
-  :host([data-component-name="VariantManagementManageViewsDialog"]) .ui5-popup-content{
-    padding: 0;
+const styles = {
+  ManageViewsDialog: {
```
Lukas742, line 28: Please use the camelCase notation for JSS style classes.

Lukas742 (changes requested):

Lukas742 (approved):

PR #2932, opened 2022-05-10
fix(Modals): support React 18 by rendering Portals instead of creating new roots

### LINKED ISSUE #2730: Modals: support react 18
The [`Modals`](https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/main/src/components/Modals/index.tsx) helper class is currently the only class which is using `react-dom` APIs directly.

Currently, we don't support the new React 18 features because we're still calling `ReactDOM.render` to open the modals. 
This should be refactored to use `ReactDOMClient.createRoot()`.

Challenge:
Is there a way of keeping supporting both react 16/17 and 18?

Options:
1. Keep the current API, but the rendered modals won't support any new React 18 feature
2. Drop support for react < 18. This would allow us using the `react-dom/client` module.
3. Is there any way of checking the react-dom version and then load `react-dom/client` on demand?
    I've played around with such a pattern:
    ```js
    import * as ReactDOM from 'react-dom';
   
   let createRoot;
    if ('createRoot' in ReactDOM') {
      createRoot = require(/* webpackIgnore: true */ 'react-dom/client').createRoot;
    }
    ```

MarcusNotheis:
BREAKING CHANGE: In order to use the `Modals` helper, you **must** wrap your Application into the `<ThemeProvider />` component. For most users this should not be a breaking change as the `<ThemeProvider />` was required for other purposes already.

Fixes #2730


packages/main/src/components/Modals/Modals.stories.mdx
```diff
@@ -90,15 +116,20 @@ The `showDialog` method returns an object with the following properties:
...
 
 **Parameters**
 
-| Parameter     | Description                                                                                       |
-| ------------- | ------------------------------------------------------------------------------------------------- |
-| `props`       | All supported `Popover` props (see table below). Open will always be set to `true`.               |
-| _`container`_ | Optional container where the `Popover` should be mounted. Defaults to a div appended to the body. |
+| Parameter     | Description                                                                            |
+| ------------- | -------------------------------------------------------------------------------------- |
+| `props`       | All supported `Popover` props (see table below). Open will always be set to `true`.    |
```
Lukas742, line 131:
```suggestion
| `props`       | All supported `Popover` props (see table below).`open` will always be set to `true`.    |
```

packages/main/src/components/Modals/Modals.stories.mdx
```diff
@@ -141,15 +173,20 @@ The `showPopover` method returns an object with the following properties:
...
 
 **Parameters**
 
-| Parameter     | Description                                                                                                 |
-| ------------- | ----------------------------------------------------------------------------------------------------------- |
-| `props`       | All supported `ResponsivePopover` props (see table below). Open will always be set to `true`.               |
-| _`container`_ | Optional container where the `ResponsivePopover` should be mounted. Defaults to a div appended to the body. |
+| Parameter     | Description                                                                                      |
+| ------------- | ------------------------------------------------------------------------------------------------ |
+| `props`       | All supported `ResponsivePopover` props (see table below). Open will always be set to `true`.    |
```
Lukas742, line 188:
```suggestion
| `props`       | All supported `ResponsivePopover` props (see table below). `open` will always be set to `true`.    |
```

packages/main/src/components/Modals/Modals.stories.mdx
```diff
@@ -190,15 +228,20 @@ The `showResponsivePopover` method returns an object with the following properti
...
 
 **Parameters**
 
-| Parameter     | Description                                                                                          |
-| ------------- | ---------------------------------------------------------------------------------------------------- |
-| `props`       | All supported `MessageBox` props (see table below). Open will always be set to `true`.               |
-| _`container`_ | Optional container where the `MessageBox` should be mounted. Defaults to a div appended to the body. |
+| Parameter     | Description                                                                               |
+| ------------- | ----------------------------------------------------------------------------------------- |
+| `props`       | All supported `MessageBox` props (see table below). Open will always be set to `true`.    |
```
Lukas742, line 243:
```suggestion
| `props`       | All supported `MessageBox` props (see table below). `open` will always be set to `true`.    |
```

packages/main/src/components/Modals/Modals.stories.mdx
```diff
@@ -39,15 +59,20 @@ import { Modals } from './index';
...
 
 **Parameters**
 
-| Parameter     | Description                                                                                      |
-| ------------- | ------------------------------------------------------------------------------------------------ |
-| `props`       | All supported `Dialog` props (see table below). Open will always be set to `true`.               |
-| _`container`_ | Optional container where the `Dialog` should be mounted. Defaults to a div appended to the body. |
+| Parameter     | Description                                                                           |
+| ------------- | ------------------------------------------------------------------------------------- |
+| `props`       | All supported `Dialog` props (see table below). Open will always be set to `true`.    |
```
Lukas742, line 74:
```suggestion
| `props`       | All supported `Dialog` props (see table below). `open` will always be set to `true`.    |
```

Lukas742 (changes requested):

packages/main/src/components/Modals/ModalsProvider.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { ModalsContext, ModalState, UpdateModalStateAction } from '../../internal/ModalsContext';
+
+export interface ModalsProviderPropTypes {
+  children: ReactNode;
+}
+
+const modalStateReducer = (state: ModalState[], action: UpdateModalStateAction) => {
+  switch (action.type) {
+    case 'set':
+      return [...state, action.payload as ModalState];
```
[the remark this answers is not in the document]
MarcusNotheis: not really because the payload of `UpdateModalStateAction` can either be a ModalState or just the ID

packages/main/src/components/ActionSheet/__snapshots__/ActionSheet.test.tsx.snap
```diff
@@ -53,6 +54,24 @@ exports[`ActionSheet Render without Crashing 1`] = `
 
 exports[`ActionSheet does not crash with other component 1`] = `
 <body>
+  Object {
```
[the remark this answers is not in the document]
MarcusNotheis: I have no idea to be honest 😅

PR #2940, opened 2022-05-12
fix(DynamicPageTitle): reduce title font-size for small devices

### LINKED ISSUE #2939: [SF][0.23.0][DynamicPage] Page title's font size does not reduce when page size reduced
Please see wiki page for reference(will provide link separately) , the page title's font size should reduce when page narrow down to small size.

Lukas742:
This PR reduces the font-size of the title when displayed on phone.

Fixes #2939


packages/main/src/components/DynamicPageTitle/index.tsx
```diff
@@ -125,11 +125,19 @@ const DynamicPageTitle = forwardRef((props: DynamicPageTitlePropTypes, ref: Ref<
   } = props as InternalProps;
 
   const classes = useStyles();
-  const containerClasses = clsx(classes.container, className, Device.isIE() && classes.iEClass);
   const [componentRef, dynamicPageTitleRef] = useSyncRef<HTMLDivElement>(ref);
   const [showNavigationInTopArea, setShowNavigationInTopArea] = useState(undefined);
   const isRtl = useIsRTL(dynamicPageTitleRef);
   const isMounted = useRef(false);
+  const [isPhone, setIsPhone] = useState(
+    Device.getCurrentRange(dynamicPageTitleRef.current?.getBoundingClientRect().width).name === 'Phone'
```
MarcusNotheis, line 133:
```suggestion
    Device.getCurrentRange(dynamicPageTitleRef.current?.getBoundingClientRect().width)?.name === 'Phone'
```
Do we need optional chaining here?

packages/main/src/components/DynamicPageTitle/index.tsx
```diff
@@ -163,6 +171,7 @@ const DynamicPageTitle = forwardRef((props: DynamicPageTitlePropTypes, ref: Ref<
         : titleContainer.borderBoxSize;
       // Safari doesn't implement `borderBoxSize`
       const titleContainerWidth = borderBoxSize?.inlineSize ?? titleContainer.target.getBoundingClientRect().width;
+      setIsPhone(Device.getCurrentRange(titleContainerWidth).name === 'Phone');
```
MarcusNotheis, line 174:
```suggestion
      setIsPhone(Device.getCurrentRange(titleContainerWidth)?.name === 'Phone');
```
Do we need optional chaining here?

MarcusNotheis (changes requested):

packages/main/src/components/DynamicPageTitle/index.tsx
Lukas742: I thought we don't, because if the parameter of `getCurrentRange` is undefined, it should fall back to the window width, but since we use optional chaining in the other implementations, I think it's saver to add it here as well. (Maybe there where issues with the implementation if the `window` wasn't defined)

MarcusNotheis (approved):

PR #2971, opened 2022-05-19
feat(OverflowToolbarButton): add option to display text of a button only in toolbar popover

### LINKED ISSUE #2900: Add an option to display the text of a button only in the toolbar popover
Currently it's not possible to render a button with only an icon inside the toolbar and add a text to the same button inside the overflow popover.

Add an [OverflowToolbarButton](https://sapui5.hana.ondemand.com/#/api/sap.m.OverflowToolbarButton) and [OverflowToolbarToggleButton](https://sapui5.hana.ondemand.com/#/api/sap.m.OverflowToolbarToggleButton) component. Use a context to tell the component whether it's rendered inside the popover or not.

alina-hs:
With the OverToolbarButton and OverflowToolbarToggleButton it is now possible to render a button with only an icon inside the toolbar and add a text to the same button inside the overflow popover of a Toolbar.

Fixes #2900


Lukas742 (changes requested):
- Please add the two buttons as `subcomponents` to the `Meta` component.

- Please create another story that covers these two buttons, you can use the same approach as for the "With overflow button" story and add a clear example of the buttons' behavior.

- Please add both components to the exports in the `index.ts` file (`packages/main/src/index.ts`)

packages/main/src/OverflowToolbarButton.ts
```diff
@@ -0,0 +1,11 @@
+import { deprecationNotice } from '@ui5/webcomponents-react-base';
```
MarcusNotheis, line 1: In my opinion, you can also delete this file because it's a new component.
These files only exist for components that have been exported via this path before.

packages/main/src/OverflowToolbarToggleButton.ts
```diff
@@ -0,0 +1,11 @@
+import { deprecationNotice } from '@ui5/webcomponents-react-base';
```
MarcusNotheis, line 1: In my opinion, you can also delete this file because it's a new component.
These files only exist for components that have been exported via this path before.

MarcusNotheis (changes requested):

packages/main/src/components/OverflowToolbarToggleButton/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * The `OverflowToolbarToggleButton` represents a toggle button that shows its text only when in the overflow area of a `Toolbar`.
+ */
+const OverflowToolbarToggleButton = forwardRef(
+  (props: OverflowToolbarToggleButtonPropTypes, ref: Ref<ToggleButtonDomRef>) => {
+    const { children, ...rest } = props;
+    const { inPopover } = useContext(OverflowPopoverContext);
+
+    return <ToggleButton {...props}>{inPopover ? children : ''}</ToggleButton>;
```
Lukas742, line 28: Only spread `rest` here and add the `ref`.
```suggestion
    return <ToggleButton {...rest} ref={ref}>{inPopover ? children : ''}</ToggleButton>;
```

packages/main/src/components/OverflowToolbarButton/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/**
+ * The `OverflowToolbarButton` represents a push button that shows its text only when in the overflow area of a `Toolbar`.
+ */
+const OverflowToolbarButton = forwardRef((props: OverflowToolbarButtonPropTypes, ref: Ref<ButtonDomRef>) => {
+  const { children, ...rest } = props;
+  const { inPopover } = useContext(OverflowPopoverContext);
+
+  return <Button {...props}>{inPopover ? children : ''}</Button>;
```
Lukas742, line 27: Only spread rest here and add the ref.
```suggestion
  return <Button {...rest} ref={ref}>{inPopover ? children : ''}</Button>;
```

Lukas742 (changes requested):

Lukas742 (approved):

PR #3224, opened 2022-07-09
chore: setup eslint

MarcusNotheis:


packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -1088,7 +1088,7 @@ const AnalyticalTable = forwardRef((props: AnalyticalTablePropTypes, ref: Ref<HT
         )}
       </div>
       <Text aria-hidden="true" id={`smartScaleModeHelper-${uniqueId}`} className={classes.hiddenSmartColMeasure}>
-        ""
+        {''}
```
Lukas742, line 1091: I think we can just self-close this `Text`, so no need for an empty string to be rendered.

packages/main/src/components/Form/index.tsx
```diff
@@ -101,6 +101,7 @@ export interface FormPropTypes extends CommonProps {
 
 const clonedChildrenForSingleColumn = (reactChildren, currentLabelSpan) =>
   React.Children.map(reactChildren, (child) => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
Lukas742, line 104: `@ts-ignore`, shouldn't be necessary here, I probably overlooked that in the initial implementation

Lukas742 (changes requested):

packages/main/src/components/AnalyticalTable/index.tsx
MarcusNotheis: children are mandatory

packages/main/src/components/Form/index.tsx
MarcusNotheis: done

Lukas742 (approved):

PR #3272, opened 2022-07-20
docs(ViewSettingDialog): update JSDoc description & types

rchrdchn:
fix(ui5-view-settings-dialog): update setConfirmedSettings param strong type from String to Object and description sample code

setConfirmedSettings' setting type should be an Object

https://github.com/SAP/ui5-webcomponents/pull/5534


packages/main/src/webComponents/ViewSettingsDialog/index.tsx
```diff
@@ -14,10 +14,10 @@ interface ViewSettingsDialogAttributes {
...
-   * Sets a JavaScript object, as settings to the `ViewSettingsDialog`. This method can be used after the dialog is initially open, as the dialog need to set its initial settings. The `ViewSettingsDialog` throws an event called "before-open", this can be used as trigger point. The object should have the following format: `{ { "sortOrder" : "Ascending", "sortBy" : "Name", "filters" : [{"Filter 1": ["Some filter 1", "Some filter 2"]}, {"Filter 2": ["Some filter 4"]}]} }`
-   * @param {string} settings - A value to be set as predefined settings.
+   * Sets a JavaScript object, as settings to the `ViewSettingsDialog`. This method can be used after the dialog is initially open, as the dialog need to set its initial settings. The `ViewSettingsDialog` throws an event called "before-open", this can be used as trigger point. The object should have the following format: `{sortOrder: "Ascending", sortBy: "Name", filters: [{Filter 1: ["Some filter 1", "Some filter 2"]}, {Filter 2: ["Some filter 4"]}]}`
+   * @param {Object} settings - A value to be set as predefined settings.
    */
-  setConfirmedSettings: (settings: string) => void;
+  setConfirmedSettings: (settings: Object) => void;
```
Lukas742, line 20: The `Object` type means "any non-nullish value", to allow any object here, use `Record<Keys, Type>`.
```suggestion
  setConfirmedSettings: (settings: Record<string, unknown>) => void;
```

packages/main/src/webComponents/ViewSettingsDialog/index.tsx
```diff
@@ -14,10 +14,10 @@ interface ViewSettingsDialogAttributes {
 
 export interface ViewSettingsDialogDomRef extends ViewSettingsDialogAttributes, Ui5DomRef {
   /**
-   * Sets a JavaScript object, as settings to the `ViewSettingsDialog`. This method can be used after the dialog is initially open, as the dialog need to set its initial settings. The `ViewSettingsDialog` throws an event called "before-open", this can be used as trigger point. The object should have the following format: `{ { "sortOrder" : "Ascending", "sortBy" : "Name", "filters" : [{"Filter 1": ["Some filter 1", "Some filter 2"]}, {"Filter 2": ["Some filter 4"]}]} }`
-   * @param {string} settings - A value to be set as predefined settings.
+   * Sets a JavaScript object, as settings to the `ViewSettingsDialog`. This method can be used after the dialog is initially open, as the dialog need to set its initial settings. The `ViewSettingsDialog` throws an event called "before-open", this can be used as trigger point. The object should have the following format: `{sortOrder: "Ascending", sortBy: "Name", filters: [{Filter 1: ["Some filter 1", "Some filter 2"]}, {Filter 2: ["Some filter 4"]}]}`
+   * @param {Object} settings - A value to be set as predefined settings.
```
Lukas742, line 18: Optional:
In my opinion, JSDoc and the description here can be improved as well. As we only adopt the description from the UI5 Web Components, this has to be adapted in their repo. Otherwise all changes will be lost when we run our script again. (e.g. if `@ui5/web-components` update their version).

That's why I'm just making a suggestion here and if you find it better that way too, you can add this suggestion to the PR in the UI5 web components repo.

```suggestion
   * @param {Object} settings - predefined settings.
   * @param {string} settings.sortOrder - sort order
   * @param {string} settings.sortBy - sort by
   * @param {Array.<Object>} settings.filters - filters
```

Lukas742 (changes requested):

packages/main/src/webComponents/ViewSettingsDialog/index.tsx
rchrdchn: @Lukas742 did you mean `Record<string, unknown>` or `Record<Keys, Type>` on the suggestion above?

Lukas742: `Record<string, unknown>` `Keys` and `Type` are just generics. See [here](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type)

MarcusNotheis (approved):

PR #3374, opened 2022-08-16
fix(SplitterLayout): adjust design to comply with guidelines

EtienneWille:

## Pull Request Checklist
- [ ] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents-react/blob/main/CONTRIBUTING.md)
- [ ] [Correct commit message style](https://github.com/SAP/ui5-webcomponents-react/blob/main/docs/Guidelines.md#commit-message-style)

### DIFF
--- a/packages/main/src/components/Splitter/index.tsx
+++ b/packages/main/src/components/Splitter/index.tsx
@@ -370,10 +370,10 @@ const Splitter = forwardRef((props: SplitterPropTypes, ref: Ref<HTMLDivElement>)
 
   const currentTheme = useCurrentTheme();
   const isHighContrast =
-    currentTheme == 'sap_fiori_3_hcb' ||
-    currentTheme == 'sap_fiori_3_hcw' ||
-    currentTheme == 'sap_horizon_hcb' ||
-    currentTheme == 'sap_horizon_hcw'
+    currentTheme === 'sap_fiori_3_hcb' ||
+    currentTheme === 'sap_fiori_3_hcw' ||
+    currentTheme === 'sap_horizon_hcb' ||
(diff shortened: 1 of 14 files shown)


packages/main/src/components/Splitter/index.tsx
```diff
@@ -343,6 +368,15 @@ const Splitter = forwardRef((props: SplitterPropTypes, ref: Ref<HTMLDivElement>)
...
   }, [isRtl, vertical]);
 
+  const currentTheme = useCurrentTheme();
+  const isHighContrast =
+    currentTheme == 'sap_fiori_3_hcb' ||
+    currentTheme == 'sap_fiori_3_hcw' ||
+    currentTheme == 'sap_horizon_hcb' ||
+    currentTheme == 'sap_horizon_hcw'
+      ? true
+      : false;
```
Lukas742, line 378: You can omit the ternary operator as the result will be `true` or `false` anyways. Also even if it shouldn't matter in this case, you should only use the equality operator (`==`) if you really need that behavior. In all other cases the strict equality operator should be used (`===`). The difference between these two is that `==` converts different types, so e.g. `0 == "0"` would be `true`. https://dorey.github.io/JavaScript-Equality-Table/
```suggestion
  const isHighContrast =
    currentTheme === 'sap_fiori_3_hcb' ||
    currentTheme === 'sap_fiori_3_hcw' ||
    currentTheme === 'sap_horizon_hcb' ||
    currentTheme === 'sap_horizon_hcw';
```

Lukas742 (changes requested):

Lukas742 (approved):

PR #3448, opened 2022-08-30
fix(ActionSheet): deprecate `alwaysShowHeader` prop

MarcusNotheis:


Lukas742 (changes requested):
Please add a note to the components JSDoc description, or extend the `header` and `headerText` description, that it will only be shown in mobile mode.

Lukas742 (approved):

PR #3541, opened 2022-09-19
fix(Toolbar): fire `onOverflowChange` event if `children` are changed

### LINKED ISSUE #3530: [SF][DynamicPageTitle][expect within two weeks]actionsToolbar onOverflowChange function not triggered after insert action
**Describe the bug**
The `actionsToolbarProps - onOverflowChange` function not triggered after insert new actions

**Isolated Example**
https://codesandbox.io/s/dynamicpage-custom-title-forked-o8e9b7?file=/src/App.js:886-905

**To Reproduce**
1. Go to https://codesandbox.io/s/dynamicpage-custom-title-forked-o8e9b7?file=/src/App.js:886-905
image

2. Click on "add action" button below the `DynamicPageTitle`
image

4. Inspect the console output
5. See the `onOverflowChange` function is not triggered after the action insert
image

**Expected behavior**
onOverflowChange function should triggered correctly after inserting new actions to DynamicPageTitle

`@ui5/webcomponents` version: 0.28.1
`@ui5/webcomponents-react` version: 1.7.1
Operating System: any
Browser: Chrome

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

Lukas742:
This PR also debounces the event, so it doesn't get called twice.

Fixes #3530

### DIFF
--- a/packages/main/src/components/Toolbar/index.tsx
+++ b/packages/main/src/components/Toolbar/index.tsx
@@ -265,21 +265,29 @@ const Toolbar = forwardRef((props: ToolbarPropTypes, ref: Ref<HTMLDivElement>) =
     [onClick, active]
   );
 
+  const prevChildren = useRef(React.Children.toArray(children));
+  const debouncedOverflowChange = useRef(debounce(onOverflowChange, 60)).current;
   useEffect(() => {
-    if (lastVisibleIndex !== null && typeof onOverflowChange === 'function') {
+    const haveChildrenChanged = prevChildren.current.length !== React.Children.toArray(children).length;
+    if ((lastVisibleIndex !== null || haveChildrenChanged) && typeof onOverflowChange === 'function') {
+      prevChildren.current = React.Children.toArray(children);
       const toolbarChildren = contentRef.current?.children;
       let toolbarElements = [];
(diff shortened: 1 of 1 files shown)


packages/main/src/components/Toolbar/index.tsx
```diff
@@ -265,21 +265,29 @@ const Toolbar = forwardRef((props: ToolbarPropTypes, ref: Ref<HTMLDivElement>) =
     [onClick, active]
   );
 
+  const prevChildren = useRef(React.Children.toArray(children));
+  const debouncedOverflowChange = useRef(debounce(onOverflowChange, 60)).current;
```
MarcusNotheis, line 269: This implies that the `onOverflowChange` prop never changes, IMO we would need an effect to update the debounced function in case the `onOverflowChange` prop is updated.

MarcusNotheis (changes requested):

packages/main/src/components/Toolbar/index.tsx
Lukas742: Good point. Fixed

MarcusNotheis (approved):

PR #3567, opened 2022-09-26
docs: create knowledge base entry covering portals

### LINKED ISSUE #3559: AnalyticalTable: "Sort column by" popover not adapting to compact mode.
**Describe the bug**
The UI's default density mode is 'Cozy', but we have set it to 'Compact' by adding `className="ui5-content-density-compact"` to our outermost component. This works fine, except for the AnalyticalTable, where when clicking on a column header, the "sort by" popover is still shown in cozy mode.
We've had the same problem with a popup which we rendered with a `ReactDOM.createPortal(popup, document.body);`, and which was therefore outside the DOM hierarchy of the app, so the setting wasn't inherited. We had to set the above className on the popup as well, which made it work. Maybe it's the same problem here.

**Isolated Example**
If it worked, here's a [sandbox](https://codesandbox.io/s/broken-firefly-dt4nk3?file=/src/App.js).
Otherwise, here's the code:
```
import React from "react";
import { AnalyticalTable, ThemeProvider } from "@ui5/webcomponents-react";

export default function App() {
  return (
    <ThemeProvider>
      <AnalyticalTable
        className="ui5-content-density-compact"
        data={[
          {
            name: "Germany",
            path: ["Germany"],
            country: undefined,
            description: undefined
          },
          {
            name: "France",
            path: ["France"],
            country: undefined,
            description: undefined
          }
        ]}
        columns={[
          { id: "name", accessor: "name", Header: "ID" },
          { id: "description", accessor: "description", Header: "Description" }
        ]}
        subRowsKey={"children"}
        style={{
          inlineSize: "100%",
          blockSize: "100%"
        }}
      />
    </ThemeProvider>
  );
}
```

**To Reproduce**
Click on a column header of the table

**Expected behavior**
The popover should be in compact mode, but it's in cozy.

**Screenshots**
Let me know if you need one.

@ui5/webcomponents version: "1.7.1"
@ui5/webcomponents-react version: "0.28.1"
Operating System: Win10
Browser: Chrome

Thanks a lot in advance!

Lukas742:
Closes #3559

### DIFF
--- a/docs/Portals.stories.mdx
+++ b/docs/Portals.stories.mdx
@@ -1,0 +1,66 @@
+import { Meta } from '@storybook/addon-docs';
+import { TableOfContent } from '../.storybook/components/TableOfContent';
+import { Footer } from '@docs/Footer';
+
+<Meta title="Knowledge Base/Working with Portals" />
+
+# Working with Portals
+
+This entry explains why portals are used in UI5 Web Components for React components and what you need to consider when using them.
+
+<br />
+
+<TableOfContent headingSelector="h2" />
+
+<br />
+
+## Why Portals?
+
+In most cases when a React component like the `AnalyticalTable` implements a popover like a `Dialog` or `ResponsivePopover`, we use a [React Portal](https://reactjs.org/docs/portals.html) to mount it outside the parents DOM hierarchy, to prevent unwanted side effects (wrong positioning, alignment, etc.).
+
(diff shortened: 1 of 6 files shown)


MarcusNotheis (approved):
LGTM. Should we add a Link to this page to all `portalContainer` props as well?

PR #3619, opened 2022-10-07
feat(charts): allow full customization of reference lines

Lukas742:
This PR adds the possibility to add all available `ReferenceLine` props of `recharts` to each reference line.

It also sorted all imports and executed prettier on all files, that's why there are that many changes. The main change is in each chart's `tsx` file.


MarcusNotheis (approved):
LGTM. One thing:
Can you please revert the changes in the CHANGELOG and the README? The new line breaks looks somehow strange.

PR #3921, opened 2022-12-27
feat: TimelineChart

### LINKED ISSUE #628: Gantt Component
Typical Gantt Component

NPKompleet:
This commit helps to add a TimelineChart. This chart can also be used to make Gantt Chart.
Closes #628 

## Pull Request Checklist


Lukas742:
Hi @NPKompleet 

thanks also from me for the PR :)

since some topics have piled up over the holiday season and some colleagues are still on vacation, we won't get around reviewing the PR until next week.

NPKompleet:
Hi @Lukas742,

You're welcome. Please, take all the time you need. I understand that it might take a while to review my contribution. In the meantime, if I see another interesting component that requires contribution in the issues list, I might just pick it up whenever I have free time to devote to it.

Lukas742 (changes requested):
Review pt.1 (pt.2 is coming next week)

I only reviewed part of the changes and concentrated more on structure and consistency, than on logic and implementation details.

Here are som general things I did find:
- only use inline style when absolutely necessary.
- don't use inline functions
- use ES6 default values to define default values for props (we will refactor our components to omit `defaultProps` as they may not be supported anymore with future React versions)
- each publicly exposed component has to accept a `ref`, you can use `forwardRef` for this

NPKompleet:
Hi @Lukas742,
Thanks for the review so far. I will go through them and try to implement your suggestions later this week.

packages/charts/src/components/TimelineChart/TimelineChartAnnotation.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const style: CSSProperties = {
+    width: width != null ? `${width}px` : 'auto',
+    height: height != null ? height : rowHeight,
+    position: 'absolute',
+    top: `${rowIndex * rowHeight}px`
+  };
+
+  return (
+    <div className="timeline-chart-annotation" style={style}>
```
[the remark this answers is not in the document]
NPKompleet: Hi @Lukas742,

I have started working on the suggestions in the review 2 days ago. The only one that I still don't have a good solution for are these classes I added to some of the subcomponents. I am using the classes as a querySelector in the tests to be able to check the behavior of the component under certain conditions. I suppose I could just add a data-testId and use a getByTestId in the test too but that is probably the same thing. Do you have an alternative idea? Thanks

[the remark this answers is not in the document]
Lukas742: First of all, sorry that I haven't continued my review. It has been a pretty busy week. I'll try to get to it in the beginning of next week.

We usually don't add attributes or classes just for testing purposes to our components. In this case it should be sufficient to add e.g. a `data-testid` data attribute to the `TimelineChartAnnotation` component that is rendered in the test env, and then you'll be able to query it by using [getByTestId](https://testing-library.com/docs/react-testing-library/cheatsheet) of react-testing-library.

If an element is not exposed at the top level and data attributes cannot be added via React props, then we usually add a `data-component-name` attribute, describing the element. `component-name` is actually the wrong wording for this, but since this is used in many components, we sticked to it to stay consistent. You can search our project for this attribute to see how we usually name elements like this. [Example1](https://github.com/SAP/ui5-webcomponents-react/blob/dfbf91fb384313d224d618079268f6abcd73208a/packages/main/src/components/AnalyticalTable/TableBody/VirtualTableBody.tsx#L97), [Example2](https://github.com/SAP/ui5-webcomponents-react/blob/dfbf91fb384313d224d618079268f6abcd73208a/packages/main/src/components/AnalyticalTable/TableBody/VirtualTableBodyContainer.tsx#L94)

packages/charts/src/components/TimelineChart/TimelineChart.tsx
```diff
@@ -0,0 +1,341 @@
...
+        className="timeline-chartbody-container"
+        ref={bodyConRef}
+        style={{
+          width: unscaledBodyWidth,
+          height: height,
+          overflowX: 'hidden',
+          overflowY: 'hidden',
+          cursor: getCursor()
+        }}
+        onMouseDown={(e) => onMouseDown(e)}
```
Lukas742, line 265:
```suggestion
        onMouseDown={onMouseDown}
```

Lukas742 (changes requested):
__Review pt.2__

We would like to introduce this component as experimental. Could you please add a note to the JSDocs description of the component and highlight it in the documentation as well? E.g:
> __Experimental Component!__
> This component is experimental and not subject to semantic versioning. Therefore, you could face breaking changes when updating versions. Please use with caution!
___
Currently there are no callbacks available when e.g. clicking on a task. This is fine for me and doesn't need to be implemented in the initial draft, but I found it worth mentioning.
___
In my opinion inequality comparison (`!=`) is hard to read, if you want to explicitly exclude for example `undefined` and `null` I find it easier to read if both are just excluded by using the strict inequality operator and often a check for falsy values is sufficient as well. @MarcusNotheis how do you feel about this?
___
Additionally to a ref, all public components should be stylable, meaning they should support props of the [CommonProps](https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/main/src/interfaces/CommonProps.ts) interface. They should also accept all html attributes of the respective top level element. 
___
Please don't use inline functions. In most cases we try to avoid inlining functions, because of readability.

NPKompleet:
Hi @Lukas742,

I have finished adding all the changes you requested. Please let me know if there is any more change or addition you would like

Lukas742:
Thanks again :) 
I will try to finish my review next week

packages/charts/src/components/TimelineChart/chartbody/TimelineConnections.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+/**
+ * This holds all the arrows that show the connections between different tasks.
+ */
+const TimelineChartConnections = ({ dataSet, width, rowHeight, bodyRect }: TimelineChartConnectionsProps) => {
+  const [connectionDataState, setConnectionDataState] = useState<ConnectionData[]>([]);
+  useLayoutEffect(() => {
+    const connectionData: ConnectionData[] = [];
+
+    const generateConnectionData = (activities: ITimelineChartTask[] | ITimelineChartMileStone[]) => {
```
Lukas742, line 52: Can this function be moved outside of the component, so it doesn't get recreated every time the effect is executed?

NPKompleet:
The latest request changes have now been added

packages/charts/src/components/TimelineChart/TimelineChart.stories.mdx
```diff
@@ -0,0 +1,321 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
+import { DocsHeader } from '@docs/DocsHeader';
+import { Footer } from '@docs/Footer';
+import { TimelineChart } from './TimelineChart';
+import { TimelineChartAnnotation } from './TimelineChartAnnotation';
+import { dummyDataSet, dummyDiscreteDataSet, schedulingEDFData, inventionDataset } from './examples/Dataset.ts';
+import { TimingFigure, Invention } from './examples/Annotations.tsx';
+import SubcomponentsSection from '../../../../../.storybook/docs/SubcomponentsSection.md';
```
MarcusNotheis, line 8:
```suggestion
import SubcomponentsSection from '@sb/docs/SubcomponentsSection.md?raw';
```

packages/charts/src/components/TimelineChart/TimelineChart.stories.mdx
```diff
@@ -0,0 +1,321 @@
+import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
```
MarcusNotheis, line 1:
```suggestion
import { ArgsTable, Canvas, Markdown, Meta, Story } from '@storybook/addon-docs';
```

packages/charts/src/components/TimelineChart/TimelineChart.stories.mdx
```diff
@@ -0,0 +1,321 @@
...
+
+### Loading Placeholder
+
+<Canvas>
+  <Story name="Loading Placeholder">
+    <TimelineChart />
+  </Story>
+</Canvas>
+
+<SubcomponentsSection />
```
MarcusNotheis, line 314: <Markdown>{SubcomponentsSection}</Markdown>

Lukas742 (approved):

MarcusNotheis (approved):
Thanks for your contribution, looks good to me!

PR #3934, opened 2023-01-03
fix(Toolbar): fire onClick on Space/Enter press, fix info bar colors

### LINKED ISSUE #3914: <Toolbar/> Background color is not matching as design in Morning Horizon Theme
**Describe the bug**

Toolbar where Active is True and Design is Info, 
Below is the code snippet. 
```
<Toolbar design={ToolbarDesign.Info} toolbarStyle={ToolbarStyle.Standard} active>
{/*Content*/}
</Toolbar>
```
With this as per the design wiki the toolbar should suppose to have this background `--sapInfobar_Active_Background;` if the theme is **Morning Horizon**. 

Observation is `--sapInfobar_Active_Background;` color is there but with `undefined` color.

A screenshot is taken from the design wiki:

Screenshot 2022-12-24 at 2 46 28 PM

Screenshot 2022-12-24 at 2 46 20 PM

But with the same properties, I tried with storybook, and it's not having the same styling as per the design wiki defined. 

Storybook Link:
https://sap.github.io/ui5-webcomponents-react/?path=/story/layouts-floorplans-toolbar--default-story&args=design:Info;active:true

These colors (and related CSS variables) look correct for Quartz Dark w.r.t. Fiori 3 but these colors (and CSS variables) don't look correct for **Morning Horizon**. 

**Isolated Example**
(https://codesandbox.io/s/ui5-webcomponents-react-template-29l34).*](https://sap.github.io/ui5-webcomponents-react/?path=/story/layouts-floorplans-toolbar--default-story&args=design:Info;active:true)

**To Reproduce**
Go To:
https://sap.github.io/ui5-webcomponents-react/?path=/story/layouts-floorplans-toolbar--default-story&args=design:Info;active:true

Click On Canvas:
Try changing different Theme from  Header Toolbar 

**Expected behavior**
Color Should be as per deisign wiki

**Screenshots**
This is what we can see after implementation 
Screenshot 2022-12-24 at 1 01 01 PM

"@ui5/webcomponents": "1.8.0",
"@ui5/webcomponents-fiori": "1.8.0",
"@ui5/webcomponents-react": "1.3.1",
"@xweb/common-messagebundles": "^1.0.8",
"@xweb/common-messages": "3.3.0",
"@xweb/common-react": "3.3.7",

Operating System: MAC OS 
Browser: Chrome

Lukas742:
Fixes #3914

### DIFF
--- a/packages/main/src/components/Toolbar/index.tsx
+++ b/packages/main/src/components/Toolbar/index.tsx
@@ -1,6 +1,7 @@
 import {
   debounce,
   enrichEventWithDetails,
+  useCurrentTheme,
   useI18nBundle,
   useIsomorphicLayoutEffect,
   useSyncRef
@@ -87,7 +88,7 @@ export interface ToolbarPropTypes extends Omit<CommonProps, 'onClick' | 'childre
    */
   overflowPopoverRef?: Ref<PopoverDomRef>;
   /**
-   * Fired when the user clicks on the `Toolbar`, if the `active` prop is set to "true".
+   * Fired if the `active` prop is set to true and the user clicks or presses Enter/Space on the `Toolbar`.
    */
   onClick?: (event: CustomEvent) => void;
   /**
@@ -135,6 +136,7 @@ const Toolbar = forwardRef<HTMLDivElement, ToolbarPropTypes>((props, ref) => {
   const overflowContentRef = useRef(null);
   const overflowBtnRef = useRef(null);
   const [minWidth, setMinWidth] = useState('0');
+  const currentTheme = useCurrentTheme();
 
   const i18nBundle = useI18nBundle('@ui5/webcomponents-react');
   const showMoreText = i18nBundle.getText(SHOW_MORE);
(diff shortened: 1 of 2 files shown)


packages/main/src/components/Toolbar/index.tsx
```diff
@@ -297,10 +300,17 @@ const Toolbar = forwardRef<HTMLDivElement, ToolbarPropTypes>((props, ref) => {
   return (
     <CustomTag
       style={styleWithMinWidth}
-      className={clsx(toolbarClasses, overflowNeeded && classes.hasOverflow)}
+      className={clsx(
+        toolbarClasses,
+        overflowNeeded && classes.hasOverflow,
+        (currentTheme === 'sap_horizon' || currentTheme === 'sap_horizon_dark') && 'shadowOutline'
```
MarcusNotheis, line 306: I don't think we should read the current theme via JS here. Using the theme dependant CSS Variables would be the better fit in my opinion.

MarcusNotheis (changes requested):

MarcusNotheis (approved):

saorabhkr:
Hi @Lukas742, 
Any idea when it will available for use?
Best Regards,

Lukas742:
Release is on its way :)

PR #4004, opened 2023-01-19
refactor(TypeScript): replace all public `ReactElement` types with `ReactNode`

Lukas742:
`ReactElement` has no real advantage over `ReactNode` and makes it even harder to type some structures of children.

```tsx
export interface ObjectPageSectionPropTypes {
  id: string;
}

interface ObjectPagePropTypes {
  children?: ReactElement<ObjectPageSectionPropTypes> | ReactElement<ObjectPageSectionPropTypes>[];
}

const TestComp = (props: ObjectPagePropTypes) => {
  return <>{props.children}</>
}

const TestComp2 = (props: ObjectPageSectionPropTypes) => {
  return <></>
}

const TestComp3 = () => {
  return (
    <>
      <TestComp>
      {/* shouldn't throw ts-error, but does*/}
        <TestComp2 id="1" />
        {[<TestComp2 id="2" />,<TestComp2 id="3" />].map(item=> item)}
      </TestComp>
      <TestComp>
      {/* should throw ts-error, but doesn't */}
        <div />
      </TestComp>
    </>
  );
}
```

### DIFF
--- a/packages/main/src/components/ActionSheet/index.tsx
+++ b/packages/main/src/components/ActionSheet/index.tsx
@@ -1,7 +1,7 @@
 import { isPhone } from '@ui5/webcomponents-base/dist/Device.js';
 import { useI18nBundle, useSyncRef } from '@ui5/webcomponents-react-base';
 import clsx from 'clsx';
-import React, { Children, forwardRef, ReactElement, ReactNode, useReducer, useRef } from 'react';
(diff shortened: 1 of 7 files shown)


MarcusNotheis (approved):

Lukas742:
This PR would introduce breaking TypeScript changes.

PR #4049, opened 2023-01-27
test: add cypress commands and queries

Lukas742:
We should keep this PR open and add Cypress commands or queries to it as the need for a new one arises.


Lukas742:
superseded by https://github.com/SAP/ui5-webcomponents-react/pull/4289

PR #4074, opened 2023-02-01
feat(charts): allow passing props to all placeholder components

Lukas742:


MarcusNotheis (changes requested):
I would like to get rid of the `react-content-loader` dependency and rather build the loading placeholder using a plain css animation.
What do you think?

Lukas742:
> I would like to get rid of the `react-content-loader` dependency and rather build the loading placeholder using a plain css animation. What do you think?

Yup, let's do that.

Closed.

PR #4078, opened 2023-02-02
test(Loader): migrate Jest tests to Cypress

Lukas742:


packages/main/src/components/Loader/Loader.cy.tsx
```diff
@@ -0,0 +1,32 @@
...
+      .should('have.css', 'animation-duration', '0s')
+      .should('have.css', 'background-size', '50%');
+  });
+  it('with delay', () => {
+    cy.mount(<Loader delay={300} data-testid="loader" />);
+    cy.findByTestId('loader', { timeout: 200 }).should('not.exist');
+    cy.findByTestId('loader').should('be.visible');
+  });
+
+  //todo pass through props test
```
MarcusNotheis, line 31: Why no pass through props test here?

MarcusNotheis (approved):

packages/main/src/components/Loader/Loader.cy.tsx
Lukas742: It's not yet available on the main branch, but only [here](https://github.com/SAP/ui5-webcomponents-react/pull/4040/files#diff-19071bcf809aa76dfe0f43ab79ea831f5ebe247ecb2879eea51774c6c32f87b5).

The comment is there so we don't forget to add it to all tests afters it's available.

PR #4098, opened 2023-02-06
docs: explain how to open popovers by reference

### LINKED ISSUE #4096: Popover: Allow HTMLElement as opener
The `opener` prop of the [`Popover` component](https://sap.github.io/ui5-webcomponents-react/?path=/docs/modals-popovers-popover--default-story#properties) only accepts a string ID of the respective element.
In our use case, we have a list of an unknown number of elements and want to display a popover on each element. This would require us to generate unique IDs for each element, which isn't trivial to do in React < 18 where the useId hook doesn't exist yet. I've seen workarounds where an `<IdContextProvider>` is wrapped around the application that keeps track of how many unique Ids it has generated and increases the counter by 1 everytime a new one is requested. Before resorting to a such contraption, however, I wanted to ask if it was possible to make `opener` accept an HTMLElement, the way the opener prop of the [Menu component ](https://sap.github.io/ui5-webcomponents-react/iframe.html?viewMode=docs&id=modals-popovers-menu--default-story&args=#properties) does.

Thanks a lot in advance!

Lukas742:
Closes #4096

### DIFF
--- a/packages/main/src/components/ActionSheet/ActionSheet.stories.mdx
+++ b/packages/main/src/components/ActionSheet/ActionSheet.stories.mdx
@@ -6,7 +6,7 @@ import { useState } from 'react';
 import { PopoverHorizontalAlign } from '../../enums/PopoverHorizontalAlign';
 import { PopoverPlacementType } from '../../enums/PopoverPlacementType';
 import { PopoverVerticalAlign } from '../../enums/PopoverVerticalAlign';
-import { Button } from '../../webComponents/Button';
+import { Button, MessageStrip } from '../../webComponents';
 import { ActionSheet } from './index';
 import { isChromatic } from '../../../../../.storybook/utils';
 
@@ -121,6 +121,37 @@ const MyComponentWithActionSheet = () => {
           setActionSheetIsOpen(false);
         }}
       />
+    </>
+  );
+};
+```
+
+**Opening an `ActionSheet` by reference and not by `id`**
+
+The web component exposes a way to pass a reference of an element instead of an `id` to the `opener` prop. Since this is not supported when passing the reference in a declarative way to a React `prop`, you have to attach the ref directly on the web component.
+You can do that by e.g. leveraging a React Ref, and then set the corresponding property there.
+
+<MessageStrip hideCloseButton>
+  Changing properties outside of the React rendering cycle can lead to unwanted behavior, please be extra cautious when
+  using this approach!
+</MessageStrip>
+
+```jsx
+const ActionSheetComponent = () => {
+  const popoverRef = useRef(null);
+  const [open, setOpen] = useState(false);
+  const handleOpenerClick = (e) => {
+    if (popoverRef.current) {
+      popoverRef.current.opener = e.target;
+      setOpen((prev) => !prev);
+    }
+  };
+  return (
+    <>
+      <Button onClick={handleOpenerClick}>Opener</Button>
+      <ActionSheet ref={popoverRef} open={open}>
+        {/* Content */}
+      </ActionSheet>
     </>
   );
 };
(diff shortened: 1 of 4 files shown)


packages/main/src/webComponents/ResponsivePopover/ResponsivePopover.stories.mdx
```diff
@@ -116,6 +117,37 @@ const MyComponentWithPopover = () => {
 };
 ```
 
+**Opening a `ResponsivePopover` by reference and not by `id`**
+
+The web component exposes a way to pass a reference of an element instead of an `id` to the `opener` prop. Since this is not supported when passing the reference in a declarative way to a React `prop`, you have to attach the ref directly on the web component.
```
MarcusNotheis, line 122:
```suggestion
This web component exposes a way to pass a reference of an element instead of an `id` to the `opener` prop. Since this is not supported when passing the reference in a declarative way to a React `prop`, you have to attach the ref directly on the web component.
```

packages/main/src/webComponents/Popover/Popover.stories.mdx
```diff
@@ -112,6 +112,37 @@ const MyComponentWithPopover = () => {
 };
 ```
 
+**Opening a `Popover` by reference and not by `id`**
+
+The web component exposes a way to pass a reference of an element instead of an `id` to the `opener` prop. Since this is not supported when passing the reference in a declarative way to a React `prop`, you have to attach the ref directly on the web component.
```
MarcusNotheis, line 117:
```suggestion
This web component exposes a way to pass a reference of an element instead of an `id` to the `opener` prop. Since this is not supported when passing the reference in a declarative way to a React `prop`, you have to attach the ref directly on the web component.
```

packages/main/src/components/ActionSheet/ActionSheet.stories.mdx
```diff
@@ -126,6 +126,37 @@ const MyComponentWithActionSheet = () => {
 };
 ```
 
+**Opening an `ActionSheet` by reference and not by `id`**
+
+The web component exposes a way to pass a reference of an element instead of an `id` to the `opener` prop. Since this is not supported when passing the reference in a declarative way to a React `prop`, you have to attach the ref directly on the web component.
```
MarcusNotheis, line 131:
```suggestion
The underlying `ui5-responsive-popover` web component exposes a way to pass a reference of an element instead of an `id` to the `opener` prop. Since this is not supported when passing the reference in a declarative way to a React `prop`, you have to attach the ref directly on the web component.
```

packages/main/src/webComponents/Menu/Menu.stories.mdx
```diff
@@ -102,6 +102,37 @@ const MyComponentWithMenu = () => {
 };
 ```
 
+**Opening a `Menu` by reference and not by `id`**
+
+The web component exposes a way to pass a reference of an element instead of an `id` to the `opener` prop. Since this is not supported when passing the reference in a declarative way to a React `prop`, you have to attach the ref directly on the web component.
```
MarcusNotheis, line 107:
```suggestion
The underlying `ui5-responsive-popover` web component exposes a way to pass a reference of an element instead of an `id` to the `opener` prop. Since this is not supported when passing the reference in a declarative way to a React `prop`, you have to attach the ref directly on the web component.
```

MarcusNotheis (changes requested):
The phrase `The web component exposes` was a bit unclear to me, left some suggestions.

packages/main/src/webComponents/ResponsivePopover/ResponsivePopover.stories.mdx
Lukas742: You're right, this is better. Done :)

MarcusNotheis (approved):

PR #4101, opened 2023-02-06
refactor: always use named icon import

MarcusNotheis:


Lukas742 (approved):

packages/main/src/components/Toolbar/OverflowPopover.tsx
```diff
@@ -1,5 +1,5 @@
-import overflowIcon from '@ui5/webcomponents-icons/dist/overflow.js';
-import { Device, useSyncRef } from '@ui5/webcomponents-react-base';
```
Lukas742, line 2: Please add `useSyncRef` again

Lukas742 (changes requested):

packages/main/src/components/Toolbar/OverflowPopover.tsx
MarcusNotheis: done, thanks

Lukas742 (approved):

Issue #4091, opened 2023-02-05
Support Server Side Rendering


MarcusNotheis:
**Is your feature request related to a problem? Please describe.**
It should be possible to use UI5 Web Components (for React) in static site generation frameworks like Next.js.

**Steps**
- [ ] create a `ssr` build folder where all static web components imports are removed and the web components are loaded on the client via `useEffect`. 
   _This is not a real server side rendering of the Web Components as we miss the ShadowDOM part here. The server will just send a `ui5-button` tag to the client where it get's hydrated to a web component._
- [  ]Find a way how to deal with icon imports
- [ ] Move any client-only code to `useEffect` or make sure that it is never executed on the server. (Checks for `typeof window !== 'undefined'`).
- [ ] Make all used files of `@ui5/webcomponents-base` SSR ready (they should not throw an error on the server):
   - [ ] `Device`
   - [ ] `MediaRange`
   - [ ] `EventProvider`
   - [ ] `i18nBundle`
   - [ ] `Theming`
   - [ ] `config/Theme`
   - [ ] `config/RTL`
   - [ ] `config/Language`
   - [ ] `theming/ThemeLoaded`
   - [ ] `locale/languageChange`
   - [ ] `Boot`
   - [ ] `CustomElementsScope`

sense-net:
Hello Marcus, can we start development with nextjs 13 ssr and rsc, as i understand , 3 issue not ready  
      Find a way how to deal with icon imports
      https://github.com/SAP/ui5-webcomponents/issues/6666
      Mark packages as type module so that transpileModules doesn't need to be configured
 what does this mean exactly , thanks

MarcusNotheis:
Released in 1.17.1 🎉 

https://twitter.com/marcusnotheis/status/1676163374080569347

PR #4116, opened 2023-02-08
refactor(MessageBox): make MessageBox SSR ready

MarcusNotheis:
and rewrite test to cypress

Part of #4091


Lukas742 (changes requested):
Please update Jest snapshots

Lukas742 (approved):

PR #4155, opened 2023-02-15
docs: update to storybook 7 beta

MarcusNotheis:


Lukas742 (changes requested):
- [ ] There are no anchors available in docs mode - https://github.com/storybookjs/storybook/issues/20519
- [x] TOC is too large when rendered as sidebar, making it overflow for components with many headings (e.g. AnalyticalTable) and navigation is broken, because of missing anchors
- [ ] Many components are displaying an empty args table in docs mode
- [ ] The storybook configuration popovers (Theme, Content Density, etc.) are not closed on selection anymore, you have to press the button again to close it
- [x] Plain text is always wrapped inside styled `p` tags, even when inside components, resulting in the MessageStrip looking weird (especially with links). A workaround for this is passing the text to the `children` prop, or creating a custom component.
- [ ] Footer of docs is mounted too far up for some components (e.g. RatingIndicator) (shouldn't be a problem once args table is enabled again)

Broken stories:
- [x] FCL (docs)
- [x] NotificationListGroupItem (docs)
- [x] all date pickers (freezes browser tab (oom))
- [x] Shellbar (oom)
- [x] Popover (docs)
- [x] ResponsivePopover (docs)
- [x] IllustratedMessage (docs)

PR #4289, opened 2023-03-10
feat: introduce new `@ui5/webcomponents-cypress-commands` package

Lukas742:


package.json
```diff
@@ -26,7 +26,8 @@
     "examples:recreate-seed": "rimraf examples/seed-test && npx create-react-app examples/seed-test --template file:./packages/cra-template-seed",
     "examples:start-seed": "cd examples/seed-test && SKIP_PREFLIGHT_CHECK=true yarn start",
     "chromatic": "cross-env STORYBOOK_ENV=chromatic npx chromatic --build-script-name build:storybook",
-    "postinstall": "rimraf node_modules/@types/mocha && husky install && lerna run build:i18n"
+    "postinstall": "rimraf node_modules/@types/mocha && husky install && lerna run build:i18n",
+    "create-cypress-commands-docs": "typedoc --plugin typedoc-plugin-markdown --entryPointStrategy expand --excludePrivate --out ./packages/cypress-commands/docs ./packages/cypress-commands --json ./packages/cypress-commands/api.json"
```
MarcusNotheis, line 30: Instead of specifing all options inline, I would prefer a `typedoc.json` with the following content:
```json
{
  "$schema": "https://typedoc.org/schema.json",
  "entryPoints": ["packages/cypress-commands"],
  "entryPointStrategy": "expand",
  "out": "temp/typedoc",
  "plugin": ["typedoc-plugin-markdown"],
  "exclude": [
    "**/dist/**/*",
    "**/test/**/*"
  ],
  "excludePrivate": true
}
```
This line can then be changed to:
```suggestion
    "create-cypress-commands-docs": "typedoc --json ./packages/cypress-commands/api.json && rimraf temp/typedoc"
```

MarcusNotheis (changes requested):

package.json
Lukas742: I also moved generating the `json` file into the `typedoc` json

Lukas742:
The `check-bundle-size` action is only failing, because the `clean:remove-modules` script is not yet available on the target branch.

PR #4304, opened 2023-03-13
docs: migrate List and all list items to CSF-3

MarcusNotheis:


packages/main/src/webComponents/List/List.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+<Markdown>{ListDescription}</Markdown>
+
+<br />
+
+<Markdown>{SubcomponentsSection}</Markdown>
+
+## StandardListItem
+
+<Description of={StandardListItem} />
```
Lukas742, line 36: The description is already included in the imported markdown description below. Either we have do adjust the markdown description here, or show the whole description (usage, parts section, etc.) on top of the args table.

The same goes for all web-component sub-components.

Lukas742 (changes requested):

packages/main/src/webComponents/List/List.mdx
MarcusNotheis: updated.

Lukas742 (approved):

PR #4314, opened 2023-03-13
fix(MessageItem): correct style for `"None"` message icon, make `type` and `children` optional (TypeScript)

Lukas742:
This PR also migrates the `MessageView` story to the CSF.


packages/main/src/components/MessageView/MessageView.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+    return (
+      <>
+        <MessageViewButton
+          counter={3}
+          type={ValueState.Error}
+          onClick={(e) => {
+            ref.current.showAt(e.target);
+          }}
+        />
+        <ResponsivePopover
```
MarcusNotheis, line 184: we need `createPortal` here, otherwise the popover is not correctly displayed.

MarcusNotheis (changes requested):
Can you also add an overall summary to the `MessageView` component? Something like:
```js
/**
 * The `MessageView` is used to display a summarized list of different types of messages (error, warning, success, and information messages).
 */
```

MarcusNotheis (approved):

PR #4377, opened 2023-03-20
feat(DynamicPage): enable sticky sub-headers support

### LINKED ISSUE #3989: Sticky Subheader Slot for DynamicPage
We want to achieve a similar effect to [this](https://sapui5.hana.ondemand.com/#/entity/sap.f.DynamicPage/sample/sap.f.sample.DynamicPageWithStickySubheader). However in the DynamicPage component, there is no slot for a sticky subheader. We have a workaround implemented using `position: sticky;` with calculated `top`, but this is causing new issues with the document flow.

We would like to see a solution where we can cleanly define a component that will behave like the sticky subheader (Navbar) in the example. I do understand that this might not be the most optimal solution for the react render cycle. So maybe there is a different approach that will have the same result. 

We have tried to implement it with custom CSS. But the solution is not clean. We also had to overwrite your CSS to make the navbar fill the whole width of the page.

Lukas742:
Closes #3989

### DIFF
--- a/packages/main/src/components/DynamicPage/index.tsx
+++ b/packages/main/src/components/DynamicPage/index.tsx
@@ -11,7 +11,7 @@ import { DynamicPageAnchorBar } from '../DynamicPageAnchorBar';
 import { FlexBox } from '../FlexBox';
 import { DynamicPageCssVariables, styles } from './DynamicPage.jss';
 
-export interface DynamicPagePropTypes extends Omit<CommonProps, 'title'> {
+export interface DynamicPagePropTypes extends Omit<CommonProps, 'title' | 'children'> {
   /**
    * Determines the background color of DynamicPage.
    */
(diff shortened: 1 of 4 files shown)


packages/main/src/components/DynamicPage/index.tsx
```diff
@@ -49,8 +49,10 @@ export interface DynamicPagePropTypes extends Omit<CommonProps, 'title'> {
   footer?: ReactElement;
   /**
    * React element or node array which defines the content.
+   *
+   * __Note:__ Assigning `children` as function is recommended when implementing sticky sub-headers. You can find out more about this [here](https://sap.github.io/ui5-webcomponents-react/?path=/story/layouts-floorplans-dynamicpage--sticky-sub-headers).
    */
-  children?: ReactNode | ReactNode[];
+  children?: ReactNode | ReactNode[] | (({ stickyHeaderHeight }: { stickyHeaderHeight: number }) => ReactElement);
```
MarcusNotheis, line 55:
```suggestion
  children?: ReactNode | ReactNode[] | ((payload: { stickyHeaderHeight: number }) => ReactElement);
```

MarcusNotheis (approved):

PR #4475, opened 2023-04-11
docs: fix jsdoc handling-slots link

Lukas742:
I found it easier to just keep the unnecessary `class`es of Link components in the description, as removing them everytime is very cumbersome and adjusting the script would only make it even more complex for something that doesn't break anything.


packages/main/src/webComponents/NotificationAction/index.tsx
```diff
@@ -48,7 +48,7 @@ export interface NotificationActionPropTypes extends NotificationActionAttribute
 }
 
 /**
- * The `NotificationAction` represents an abstract action, used in the `NotificationListItem` and the `NotificationListGroupItem` items
+ * The `NotificationAction` represents an abstract action, used in the `NotificationListItem` and the `NotificationListItemBase` items
```
MarcusNotheis, line 51: This should be reverted

MarcusNotheis (changes requested):

MarcusNotheis (approved):

PR #4620, opened 2023-05-22
docs: add vite template

MarcusNotheis:


examples/vite-ts/src/App.css
```diff
@@ -0,0 +1,4 @@
+#root {
+  width: 100vw;
+  height: 100vh;
```
Lukas742, line 3: The `body` is already applying these styles. Do we need them here as well?

MarcusNotheis: nope 🙂

Lukas742 (approved):

PR #4714, opened 2023-06-11
docs: add NextJS app router example

MarcusNotheis:


examples/nextjs-app/app/todos/[id]/page.tsx
```diff
@@ -0,0 +1,49 @@
...
+    setTimeout(() => {
+      resolve(todos.at(Number(params.id)));
+    }, 500);
+  });
+
+  return (
+    <>
+      <DynamicPage showHideHeaderButton={false} headerTitle={<DynamicPageTitle header={todo?.title} />}>
+        <MessageStrip design={MessageStripDesign.Information}>
+          {`Since this is only a demo app, adjustments made here on this page won't be reflected in the todo list.`}
```
Comment (Lukas742) [inline examples/nextjs-app/app/todos/[id]/page.tsx:29]: Why use a template string here?

Lukas742 (changes requested):

examples/nextjs-app/app/todos/[id]/page.tsx
Comment (MarcusNotheis) [reply examples/nextjs-app/app/todos/[id]/page.tsx:29]: the next eslint config complains about the unescaped `'` in this line, so we would have to go either for `&apos;` or use template string to avoid this error. Using a template string sounded more convenient to me.

Comment (Lukas742) [reply examples/nextjs-app/app/todos/[id]/page.tsx:29]: You're right, I think eslint for React (or js in general) does the same.

Lukas742 (approved):

PR #4762, opened 2023-06-19
fix(TypeScript): use UI5 Web Component types in custom components

Lukas742:


MarcusNotheis (approved):
Changes look good, I would just try to make the commit message a bit more speaking then `reuse ui5wc types`, or at least not abbreviate UI5 Web Components

PR #4930, opened 2023-07-31
docs: update storybook to 7.1.1

MarcusNotheis:
Replace custom ToC with new [Storybook ToC](https://storybook.js.org/docs/react/writing-docs/autodocs#generate-a-table-of-contents)


Lukas742:
The ToC integration in Storybook is missing some features like responsiveness that our ToC currently has. So I'm closing this PR now, but we can keep the branch so we can come back to it once Storybook has improved this.

PR #4970, opened 2023-08-16
fix(ObjectPage): slightly improve scroll performance

### LINKED ISSUE #4960: ObjectPage: header is always rerendered
### Describe the bug

By scrolling though the object page, header is always re-rendered, but there is no visual change, it causes performance issues.
Calculation of changing the style of the header is triggered on scroll event, but even when style should not be changed, header is anyway re-rendered

### Isolated Example

https://codesandbox.io/p/sandbox/charming-breeze-5zv6t9?file=%2Fsrc%2FApp.tsx%3A30%2C20

### Reproduction steps

1. Create a component with `ObjectPage`
2. Provide `headerContent` and `headerTitle` props
3. add child blocks into `ObjectPage` to make it scrollable
4. Scroll up and down when header is collapsed and check re-render of component

### Expected Behaviour

Header should not be re-rendered on each scroll event. Re-render should be triggered only when header will need to change style from collapsed to expanded.

### Screenshots or Videos

Kapture 2023-08-10 at 16 10 41

Latest

Latest

### Browser

Chrome, Edge, Firefox, Safari

### Operating System

_No response_

### Additional Context

_No response_

### Relevant log output

_No response_

### Declaration

Lukas742:
This PR slightly improves the performance of the ObjectPage (a few ms).
It also refactors the Object- and DynamicPage to use logical css properties for padding.

Fixes #4960


packages/main/src/components/DynamicPage/DynamicPage.jss.ts
```diff
@@ -45,7 +45,7 @@ export const styles = {
   },
   contentContainer: {
     extend: ResponsiveContainerPadding,
-    paddingTop: '1rem',
+    paddingBlockEnd: '1rem',
```
MarcusNotheis, line 48: Is this change intended? `paddingTop` should translate to `paddingBlockStart`, shouldn't it?

packages/main/src/components/ObjectPage/index.tsx
```diff
@@ -616,18 +613,18 @@ const ObjectPage = forwardRef<HTMLDivElement, ObjectPagePropTypes>((props, ref)
 
   const renderTitleSection = useCallback(
     (inHeader = false) => {
-      const titleStyles = { ...(inHeader ? { padding: 0 } : {}), ...(headerTitle?.props?.style ?? {}) };
+      const titleInHeaderClass = inHeader ? classes.titleInHeader : undefined;
```
MarcusNotheis, line 616: What about the given `className` on the header? shouldn't we add it as well?

packages/main/src/components/DynamicPage/DynamicPage.jss.ts
Lukas742: Nope, that wasn't intended. Good catch!

packages/main/src/components/ObjectPage/index.tsx
Lukas742: Yup, forgot about that. Thanks :)

MarcusNotheis (approved):

PR #4985, opened 2023-08-21
fix(Toolbar): fix overflow popover margins

Lukas742:
- Only apply styles to immediate children in overflow popover
- Use correct margin for buttons
- Replace physical with logical CSS properties

### DIFF
--- a/packages/main/src/components/Toolbar/Toolbar.jss.ts
+++ b/packages/main/src/components/Toolbar/Toolbar.jss.ts
@@ -97,7 +97,7 @@ export const styles = {
     '& >[ui5-button]::part(button) ,& >[ui5-toggle-button]::part(button)': {
       justifyContent: 'start'
     },
-    '& >[ui5-button][icon-only]::part(button), & > [ui5-toggle-button][icon-only]::part(button)': {
+    '& >[ui5-button][icon-only]::part(button), & >[ui5-toggle-button][icon-only]::part(button)': {
       padding: 'revert'
     },
     '& >:last-child': {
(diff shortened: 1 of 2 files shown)


packages/main/src/components/Toolbar/Toolbar.jss.ts
```diff
@@ -85,23 +85,26 @@ export const styles = {
...
+    insetInlineStart: '5px !important'
   },
   popoverContent: {
     padding: CssSizeVariables.ui5WcrToolbarPopoverContentPadding,
     display: 'flex',
     flexDirection: 'column',
-    '& [ui5-toggle-button], & [ui5-button]': {
-      marginBottom: '0.25rem'
+    '& >[ui5-toggle-button], & >[ui5-button]': {
+      marginBlock: '0.25rem'
```
MarcusNotheis, line 95: Are we now using a two sided margin? or should this rather be `marginBlockEnd`?

Lukas742: Yup. Checked with SapUI5 and they do the same

packages/main/src/components/Toolbar/Toolbar.jss.ts
```diff
@@ -85,23 +85,26 @@ export const styles = {
...
+    '& >[ui5-button][icon-only]::part(button), & >[ui5-toggle-button][icon-only]::part(button)': {
       padding: 'revert'
     },
-    '& :last-child': {
-      marginBottom: 0
+    '& >:last-child': {
+      marginBlockEnd: 0
+    },
+    '& >:first-child': {
+      marginBlockStart: 0
```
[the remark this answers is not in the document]
Lukas742: I added it for consistence, but now that I think about it again, we could also just remove both, as these values aren’t in the specs anyway.

MarcusNotheis (approved):

PR #5002, opened 2023-08-24
feat(Flexbox): introduce flexbox `gap` property

marpme:
Introduces a new FlexBox property called `gap` that allows users to easily access gap functionality in a standardized way of increasing 16px (=1rem) steps.

Few examples explaining the underlying behaviour:

`gap={1}` will result in 16px gap (=1rem)
`gap={2}` will result in 32px gap (=2rem) etc.

This will ease the different mixtures of gap mechanisms with in projects using UI5.

Alternatively users can also specify full css values in the gap property as string (i.e. "0 2rem") for more advanced usages.

## Pull Request Checklist


packages/main/src/components/FlexBox/FlexBox.cy.tsx
```diff
@@ -93,6 +93,48 @@ describe('FlexBox', () => {
...
+  it('gap values', () => {
+    cy.mount(
+      <FlexBox data-testid="flexbox">
+        <span>Test 1</span>
+        <span>Test 2</span>
+        <span>Test 3</span>
+      </FlexBox>
+    );
+
+    cy.findByTestId('flexbox').should('have.css', 'gap', 'normal');
```
marpme, line 105: **Note:** [`normal`](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) should be the default value according to various sources. Therefore, this shall not cause a breaking change for existing components.

packages/main/src/components/FlexBox/FlexBox.jss.ts
```diff
@@ -1,7 +1,11 @@
+import type { CSSProperties } from 'react';
+
 export const styles = {
   flexBox: {
     display: 'flex',
 
+    gap: ({ gap }: { gap: CSSProperties['gap'] }) => (typeof gap === 'number' ? `${gap}rem` : gap),
```
marpme, line 7: **Question:** Shall we consider a different scale for numbers based styles? 

The `gap * 1rem` scaling was the easiest and most straight forward one, that I could come up with at the moment.

packages/main/src/components/FlexBox/index.tsx
```diff
@@ -27,6 +27,16 @@ export interface FlexBoxPropTypes extends CommonProps {
    * Determines whether the `FlexBox` will be sized to completely fill its container.
    */
   fitContainer?: boolean;
+  /**
+   * Defines the CSS gap between all inner children on row and column level.
+   * By default, gap values shall be provided as a number. It scales automatically in `gap * 1rem` steps.
+   *
+   * For more advanced usages patters, there's the option to provide a full CSS gap property string.
+   *
+   * <b>Note:</b> Corresponds to `gap` property of the CSS specification.
```
Lukas742, line 36:
```suggestion
   * **Note:** Corresponds to `gap` property of the CSS specification.
```

Lukas742 (changes requested):
Thank you for creating this PR :)
I've added a few notes below.

PR #5007, opened 2023-08-24
docs(Wizard): show entire code of the example

Lukas742:


packages/main/src/webComponents/Wizard/Wizard.mdx
```diff
@@ -13,7 +13,114 @@ import { excludePropsForAbstract } from '@sb/utils';
...
+    });
+    setSelected('2');
+  };
+  const goToStep3 = () => {
+    setDisabled((prev) => {
+      const { '3': _omit, ...rest } = prev;
+      return rest;
+    });
+    setSelected('3');
+  };
```
mrCherry97, line 40:
```suggestion
  const [selected, setSelected] = useState(1);

  const goToNextStep = () => {
    setSelected(selected + 1);
  };

  const goToPreviousStep = () => {
    setSelected(selected - 1);
  };
```

packages/main/src/webComponents/Wizard/Wizard.mdx
```diff
@@ -13,7 +13,114 @@ import { excludePropsForAbstract } from '@sb/utils';
...
+      setHidden((prev) => {
+        const { finalizeBtn: _omit, ...rest } = prev;
+        return rest;
+      });
+    }
+  };
+
+  return (
+    <Wizard style={{ height: '400px' }}>
+      <WizardStep icon={productIcon} titleText="Product Type" selected={selected === '1'}>
```
mrCherry97:
```suggestion
      <WizardStep icon={productIcon} titleText="Product Type" selected={selected === 1}>
```

packages/main/src/webComponents/Wizard/Wizard.mdx
```diff
@@ -13,7 +13,114 @@ import { excludePropsForAbstract } from '@sb/utils';
...
+          tempor vulputate dolor, nec dignissim quam convallis ut. Praesent vitae commodo felis, ut iaculis felis. Fusce
+          quis eleifend sapien, eget facilisis nibh. Suspendisse est velit, scelerisque ut commodo eget, dignissim quis
+          metus. Cras faucibus consequat gravida. Curabitur vitae quam felis. Phasellus ac leo eleifend, commodo tortor
+          et, varius quam. Aliquam erat volutpat
+        </Label>
+        <Button design={ButtonDesign.Emphasized} onClick={goToStep2}>
+          Step 2
+        </Button>
+      </WizardStep>
+      <WizardStep icon={hintIcon} titleText="Product Information" disabled={disabled['2']} selected={selected === '2'}>
```
mrCherry97:
```suggestion
      <WizardStep icon={hintIcon} titleText="Product Information" disabled={selected !== 2} selected={selected === 2}>
```

packages/main/src/webComponents/Wizard/Wizard.mdx
```diff
@@ -13,7 +13,114 @@ import { excludePropsForAbstract } from '@sb/utils';
...
+        </Label>
+        <CheckBox onChange={handleStep2Completed} text="Step Completed" />
+        <br />
+        {!hidden['step3Btn'] && (
+          <Button design={ButtonDesign.Emphasized} onClick={goToStep3}>
+            Step 3
+          </Button>
+        )}
+      </WizardStep>
+      <WizardStep icon={leadIcon} titleText="Pricing" disabled={disabled['3']} selected={selected === '3'}>
```
mrCherry97:
```suggestion
      <WizardStep icon={leadIcon} titleText="Pricing" disabled={selected !== 3} selected={selected === 3}>
```

mrCherry97 (commented):
Couldn't be this a little bit easier and more generative?

packages/main/src/webComponents/Wizard/Wizard.mdx
Lukas742: This approach does not provide the same behavior because you can select all steps from the beginning, which is often undesirable.

packages/main/src/webComponents/Wizard/Wizard.mdx
Lukas742: With my initial commit I forgot to update the React state when the user clicks on a step, since I'm now using a dataset to pass the respective step `string` is better here as then there is no parsing necessary.

packages/main/src/webComponents/Wizard/Wizard.mdx
mrCherry97: But you can disable `Next Step` button or bidę it, this is how I am using this, and that's the reason of my suggestion, because when you will have 10 steps, then you need 10 functions `goToXXStep` with this solution, you can have only 2 functions `goNext` and `goBack`, but you need to disable or hide navigation buttons (what you are making)

packages/main/src/webComponents/Wizard/Wizard.mdx
mrCherry97: Yes, this is connected to first suggestion, if first suggestion will be valid, then all of them are valid, if not then not

packages/main/src/webComponents/Wizard/Wizard.mdx
Lukas742: You're right, that would work, but since this is only an example and I like to keep examples simple, I'd like to stick with my approach. Thanks anyway :)

MarcusNotheis (approved):

PR #5166, opened 2023-10-17
docs: improve slot-prop note

Lukas742:


scripts/web-component-wrappers/utils.js
```diff
@@ -526,7 +526,7 @@ export const propDescription = (componentSpec, property) => {
   if (property.name !== 'children' && componentSpec?.slots?.some((item) => item.name === property.name)) {
     formattedDescription += `
           *
-          * __Note:__ This prop will be rendered as [slot](https://www.w3schools.com/tags/tag_slot.asp) (\`slot="${property.name}"\`). 
+          * __Note:__ The content of the prop will be rendered into a [&lt;slot&gt;](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) by assigning the respective global [slot](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/slot) attribute (\`slot="${property.name}"\`).
```
MarcusNotheis, line 529:
```suggestion
          * __Note:__ The content of the prop will be rendered into a [&lt;slot&gt;](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) by assigning the respective [slot](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/slot) attribute (\`slot="${property.name}"\`). 
```
Can we remove the word `global` here? Although it is technically correct, I had to read it like 5 times and was always asking myself "which global slot" 😄

MarcusNotheis (approved):

PR #5203, opened 2023-11-02
feat(AnalyticalTable): fire `onLoadMore` when resizing in `Interactive` `visibleRowCountMode`

### LINKED ISSUE #5201: [AnalyticalTable]: [VisibleRowCountMode: Interactive - Does not provide any control/event handler on resizing paginated tables]
### Describe the bug

Need any event handler for table resizer in AnalyticalTable to load next page's entries when the visibleRows exceed the page size.

### Isolated Example

Video attached

### Reproduction steps

In the screen recording attached: We have an AnalyticalTable whose data is paginated with page size =10. I have set the visibleRows={8} and visibleRowCountMode='Interactive'. So at first time when the table is loaded, we can see 8 rows (and 10 rows of data is loaded in total, so there is an overflow and we can the scroll bar).
When we scroll down then onLoadMore is called and next page's 10 entries are called. Till now things work fine.

But when during the first time table is loaded with 10 entries and we drag and increase the size of the table beyond 10 entries. Now the scroll bar is gone.

### Expected Behaviour

Page size =10 and visibleRows during first render is 8. When the table is resized and dragged down to increase the size, when 10 entries are loaded, it should fire a call just like onLoadMore to load the next set of entries.

If such a scenario is possible then please suggest ways to implement that else please let know if this can be taken as a feature request.

### Screenshots or Videos

https://github.com/SAP/ui5-webcomponents-react/assets/48648230/4ffdfd71-0470-47f4-aee2-f43ec68abcc0

0.27.0

0.27.0

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Relevant log output

_No response_

### Declaration

Lukas742:
Closes #5201 

This PR also moves the `index.tsx` types of the AnalyticalTable into a separate folder to prevent circular imports.


packages/main/src/components/AnalyticalTable/VerticalResizer.tsx
```diff
@@ -156,6 +162,14 @@ export const VerticalResizer = (props: VerticalResizerProps) => {
     };
   }, []);
 
+  const isInitial = useRef(true);
+  useEffect(() => {
+    if (!isInitial.current && rowsLength <= visibleRows) {
+      handleOnLoadMore({ type: 'custom' } as Event);
```
MarcusNotheis, line 168: `type: custom` sounds a bit generic to me, but I also have no idea on how to name it better. `tableResize` would be nice but I think we can't distinguish here whether the event was really fired due to a resize...

MarcusNotheis (approved):

PR #5205, opened 2023-11-02
feat: update to `@ui5/webcomponents@1.19.0`

Lukas742:


packages/charts/package.json
```diff
@@ -34,7 +34,7 @@
     "recharts": "2.9.2"
   },
   "peerDependencies": {
-    "@ui5/webcomponents-react": "~1.21.0",
+    "@ui5/webcomponents-react": "~1.22.0",
     "@ui5/webcomponents-react-base": "~1.20.0",
```
MarcusNotheis, line 38:
```suggestion
    "@ui5/webcomponents-react-base": "~1.22.0",
```

packages/main/src/webComponents/Input/index.tsx
```diff
@@ -106,14 +106,10 @@ export interface InputPropTypes extends InputAttributes, Omit<CommonProps, 'onCh
...
-   *    &lt;Input showSuggestions><br />
-   *    &nbsp;&nbsp;&lt;SuggestionItem text="Item #1" /><br />
-   *    &nbsp;&nbsp;&lt;SuggestionItem text="Item #2" /><br />
-   *    &lt;/Input>
-   *  </code>
-   * </pre>
+   * <Input show-suggestions>
+   *     <SuggestionItem text="Item #1"></SuggestionItem>
+   *     <SuggestionItem text="Item #2"></SuggestionItem>
+   * </Input>
```
MarcusNotheis, line 112: I think we need to encode the HTML tags here, otherwise storybook will attempt to render the input?

MarcusNotheis (changes requested):

packages/charts/package.json
Lukas742: Do we have changes in the `base` package? Otherwise the version won't be bumped, right?

MarcusNotheis: yes: https://github.com/SAP/ui5-webcomponents-react/pull/5205/files#diff-0a2e875708bc7e289f4845daec445fab53f5e136388cf365c6b7f2f7ea49b003

packages/main/src/webComponents/Input/index.tsx
Lukas742: Yup, that was an oversight after running the script. Thanks for noticing!

MarcusNotheis (approved):

PR #5328, opened 2023-12-07
fix(AnalyticalTable - TypeScript): correct `tableHooks` type

Lukas742:


packages/main/src/components/AnalyticalTable/types/index.ts
```diff
@@ -502,7 +502,7 @@ export interface AnalyticalTablePropTypes extends Omit<CommonProps, 'title'> {
   /**
    * You can use this prob to add custom hooks to the table.
```
MarcusNotheis, line 503:
```suggestion
   * You can use this prop to add custom hooks to the table.
```

MarcusNotheis (approved):

PR #5334, opened 2023-12-08
fix(AnalyticalTable): improve programmatic scroll behavior & scroll types

### LINKED ISSUE #5319: [Analytical Table] First visible row.

I don't have an easy way to specify which table row should be the first to be displayed. Sometimes I don't want the first item to be the first visible item in the table, so I need to use one of the provided methods for scrolling to a specific index, but the animation is unwanted.

I would like the firstVisibleRow property (you can check here: https://experience.sap.com/fiori-design-web/analytical-table-alv/) to be implemented, so that I can implement the expected behavior. I want to be able to choose which is the first visible row in the table.

_No response_

### Additional Context

_No response_

### Declaration

Lukas742:
Fixes #5319


packages/main/src/components/AnalyticalTable/types/index.ts
```diff
@@ -56,6 +57,12 @@ export interface ReactTableHooks {
   getResizerProps: any[];
 }
 
+export interface TriggerVertScroll {
```
MarcusNotheis, line 60: I don't understand the inferface name to be honest? Does is stand for `TriggerVerticalScroll`?

Lukas742: Yup, initially I only triggered vertical scroll via state, but as it's now for both direction I'm going to adjust the name.

MarcusNotheis (approved):

PR #5390, opened 2024-01-05
feat: update to `@ui5/webcomponents@1.21.0`

Lukas742:


packages/main/src/webComponents/Select/index.tsx
```diff
@@ -57,11 +56,12 @@ export interface SelectDomRef extends Omit<SelectAttributes, 'menu'>, Ui5DomRef
    */
   value?: string;
   /**
-   * Defines a reference (ID or DOM element) of the component's options menu, as an alternative to defining the selection's drop-down menu.
-   *
-   * **Note:** Usage of `SelectMenu` is recommended.
+   * Defines the value state of the component.
    */
-  menu?: string | HTMLElement;
```
MarcusNotheis, line 64: I think the menu getter from the dom ref can also be a HTML Element, can't it?

packages/main/src/webComponents/ToolbarSeparatorV2/index.tsx
```diff
@@ -48,7 +44,7 @@ const ToolbarSeparatorV2 = withWebComponent<ToolbarSeparatorV2PropTypes, Toolbar
 
 ToolbarSeparatorV2.displayName = 'ToolbarSeparatorV2';
 
-ToolbarSelect.defaultProps = {
+ToolbarSeparatorV2.defaultProps = {
```
MarcusNotheis, line 47: 🙈

MarcusNotheis (approved):

packages/main/src/webComponents/Select/index.tsx
Lukas742: Good catch!
I also forgot to move the `value` getter/setter to the DomRef again.

MarcusNotheis (approved):
🚢

PR #5428, opened 2024-01-18
fix: use `react-jsx` transform

Lukas742:


packages/charts/src/components/TimelineChart/TimelineChart.tsx
```diff
@@ -1,7 +1,7 @@
 import type { CommonProps } from '@ui5/webcomponents-react';
 import { throttle } from '@ui5/webcomponents-react-base';
-import type { CSSProperties, ReactNode } from 'react';
-import React, { forwardRef, useEffect, useRef, useState } from 'react';
+import { forwardRef, useEffect, useRef, useState } from 'react';
+import type * as React from 'react';
```
MarcusNotheis, line 4: I would prefer to stay in our common pattern here: `import type { CSSProperties, ReactNode } from 'react';`

packages/charts/src/components/TimelineChart/chartbody/TimelineChartRow.tsx
```diff
@@ -1,6 +1,6 @@
 import { throttle } from '@ui5/webcomponents-react-base';
-import type { CSSProperties } from 'react';
-import React, { useEffect, useRef, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
+import type * as React from 'react';
```
MarcusNotheis, line 3: Pattern: `import type { CSSProperties } from 'react';`

MarcusNotheis (changes requested):
Two minor adjustments needed to stay consistent with our common `import type` pattern.
Other than that: I would prefer to have to commit marked as at least `fix` to make sure it will be available in the release notes.

packages/charts/src/components/TimelineChart/TimelineChart.tsx
Lukas742: This is not possible here, as the component imports `MouthEvent` both from typescript and React. 

```ts
React.MouseEvent<HTMLDivElement, MouseEvent>
```

Using the named import of `CSSProperties and `ReactNode` and keeping the namespace import for `MouseEvent` also doesn't work as we would then have multiple `type` imports of the same package.

MarcusNotheis (approved):

PR #5569, opened 2024-03-04
feat(FilterBar): introduce reorder items feature

Lukas742:


packages/main/src/components/FilterGroupItem/types.ts
```diff
@@ -0,0 +1,69 @@
...
+   */
+  considerGroupName?: boolean;
+  /**
+   * Defines whether the filter is displayed with a value. If it's active an indicator will be shown in the filter configuration dialog.
+   */
+  active?: boolean;
+  /**
+   * Defines the unique id used to identify reordered filter items.
+   *
+   * __Note:__ This prop is mandatory if `enableReordering` is set to `true`.
```
MarcusNotheis, line 58:
```suggestion
   * __Note:__ This prop is mandatory if `enableReordering` is set to `true` on the `FilterBar` component.
```
Maybe we should add that `enableReordering` prop is on the FilterBar to avoid confusion

packages/main/src/components/FilterBar/FilterDialog.tsx
```diff
@@ -197,7 +222,61 @@ export const FilterDialog = (props: FilterDialogPropTypes) => {
   const hideValuesText = i18nBundle.getText(HIDE_VALUES);
   const fieldText = i18nBundle.getText(FIELD);
   const fieldsByAttributeText = i18nBundle.getText(FIELDS_BY_ATTRIBUTE);
-  const resetWarningText = i18nBundle.getText(FILTER_DIALOG_RESET_WARNING);
+
+  const visibleChildren = () =>
+    children.filter((item) => {
```
MarcusNotheis, line 227:
```suggestion
    Children.toArray(children).filter((item) => {
```
just to be on the safe side?

MarcusNotheis (approved):

packages/main/src/components/FilterBar/FilterDialog.tsx
Lukas742: This will update the `key` of the `children` which will rerender them, it also breaks the toggle filter feature, because we still use a `key` there to identify which child has changed (we should probably revise this at some point).

`.filter` is already save to use here, since the dialog always receives an array of children. I'll adjust the type for this

PR #5601, opened 2024-03-13
fix(ObjectPage): prevent initially mounting all children in `IconTabBar` mode

### LINKED ISSUE #5600: [ObjectPage]: Renders all sections regardless which one is active
### Bug Description

The **ObjectPage** renders **all** **ObjectPageSections** regardless which sections is active. In our case, we have an  ObjectPage with three section, and at the beginning the user sees only the first section. Our app breaks, because in the background it tries to render and load also the other sections which are not visible to the user.

Versions:
The bug happened after we updated to @ui5/webcomponents: ^1.22.0 and @ui5/webcomponents-react: ^1.25.1. Before it was working with versions @ui5/webcomponents: 1.21.2 and @ui5/webcomponents-react: 1.24.0.

Note: The same started happening to FlexibleColumnLayout, and this is the related [issue](https://github.com/SAP/ui5-webcomponents/issues/8434)

### Affected Component

ObjectPage, FlexibleColumnLayout

### Expected Behaviour

It should render only the visible sections as before the update.

### Isolated Example

https://stackblitz.com/edit/github-acayej

### Steps to Reproduce

Create an object page with multiple sections in tab version. Even if the first one is only visible to the user and selected, all other sections are loaded in the background.

As shown in the isolated example.

### Log Output, Stack Trace or Screenshots

Screenshot 2024-03-06 at 09 23 19

### Priority

High

@ui5/webcomponents-base: ^1.22.0 and @ui5/webcomponents-react: ^1.25.1

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

SAP Signavio

### Declaration

Lukas742:
This PR also renames some constants to improve readability. The relevant change for the fix is in line 238.

Fixes #5600

### DIFF
--- a/packages/main/src/components/ObjectPage/ObjectPage.cy.tsx
+++ b/packages/main/src/components/ObjectPage/ObjectPage.cy.tsx
@@ -1,7 +1,7 @@
 import '@ui5/webcomponents-icons/dist/AllIcons.js';
 import '@ui5/webcomponents-fiori/dist/illustrations/NoData.js';
 import type { CSSProperties } from 'react';
-import { useReducer, useRef, useState } from 'react';
+import { useEffect, useReducer, useRef, useState } from 'react';
 import type { ObjectPagePropTypes } from '../..';
 import {
   Avatar,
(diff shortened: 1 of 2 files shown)


ilhan007:
Hi Lucas  @Lukas742 

we have basically the same  issue (same behaviour as for the ObjectPageSubSection) in our repo https://github.com/SAP/ui5-webcomponents/issues/8434  filed by the same author,  but for the FCL.
Do you think it could be the same issue as this.

Lukas742:
Hi @ilhan007 

in our initial render we mounted all `children` of the component and only then filtered them for the `IconTabBar` mode. This caused the `useEffect` inside the custom component to be triggered. I can't really asses if this is the same problem, since the ObjectPage heavily relies on React syntax, but if you previously dynamically rendered columns and now always add them to the DOM and just hide them, the issue could have the same root cause. One thing I forgot to mention in the issue is that I couldn't reproduce the error in the screenshot. This PR will only fix the issue with the `useEffect`.

MarcusNotheis (approved):

PR #5701, opened 2024-04-11
refactor(AnalyticalTable): migrate from react-jss to css modules

Lukas742:


packages/main/src/components/AnalyticalTable/AnalyticalTable.module.css
```diff
@@ -0,0 +1,312 @@
...
+  position: relative;
+}
+
+.overlay {
+  position: absolute;
+  z-index: 1;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
```
MarcusNotheis, line 36: could be replaced with inset: 0?

packages/main/src/components/AnalyticalTable/AnalyticalTable.module.css
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  &[data-is-selected]:hover {
+    background-color: var(--sapList_Hover_SelectionBackground);
+  }
+}
+
+.trActive {
+  cursor: pointer;
+}
```
MarcusNotheis, line 167: this used to be `.trActive:hover` - intended change?

MarcusNotheis (approved):

packages/main/src/components/AnalyticalTable/AnalyticalTable.module.css
Lukas742: I kept it like this as it shouldn't make a difference for different writing-modes, but this way we're save if these styles are overwritten somewhere else.

packages/main/src/components/AnalyticalTable/AnalyticalTable.module.css
Lukas742: Imo it doesn't make sense setting this style only on hover as the cursor is anyway only visible when hovered.

PR #5756, opened 2024-04-25
feat(cypress-commands): add commands for dropdown menus

Lukas742:
Adds:

- `clickDropdownMenuItemByText`
- `clickDropdownMenuItem`
- `openDropDownByClick`


vigrase:
Hi!, if somebody can take a look on this will be really appreciated

rvsia:
@Lukas742 

Hello, what is the status of this PR? 😊 

We are just trying to update UI components to `1.28` and a lot of our custom commands regarding ComboBox/Select got broken - clicking on list/opening fails when a mouse is used, so we had to resort to use keyboard interactions instead. 😞

Lukas742:
Hi @vigrase @rvsia 

we will most probably release the next minor version next week, which will also include this PR. If you need this urgently, you can just copy the commands into your project until they are available in our package.

Issue #3196, opened 2022-07-05
Auto adjust width on double click between the columns


Bilal3120:
**Feature request**
When using analytical table if cell content is truncated, can we have a functionality by which the cell width gets auto adjusted when we double click in between the columns.

**Example**
<img width="785" alt="Screenshot 2022-07-05 at 2 44 56 PM" src="https://user-images.githubusercontent.com/93638146/177294837-abac869b-15f5-4486-a47d-c4ccd2e3fbb6.png">

Now when we click between the columns can the width be adjusted accordingly
<img width="785" alt="Screenshot 2022-07-05 at 2 45 02 PM" src="https://user-images.githubusercontent.com/93638146/177295519-daf7bf8c-0c61-49ca-a3d3-57c7daaf85e0.png">

Papa-Santo:
Please assign this to me. I want this feature as well, and I have the time to implement it.

My only question: If a field is not truncated, but instead has extra width, should double clicking decrease the width of the field? 

Of course it should still show all of the content but it should have only enough width to show the largest amount of text content in that field. I vote yes on this, but I understand the request is specific to making truncated fields show their full content.

MarcusNotheis:
Closed via https://github.com/SAP/ui5-webcomponents-react/pull/5758

PR #5758, opened 2024-04-26
feat(AnalyticalTable): introduce autoResize column feature (#3196)

Papa-Santo:
A functionality by which the cell width gets auto adjusted when we double click the data-resizer columns. Implemented for text.

## Pull Request Checklist

### DIFF
--- a/packages/main/src/components/AnalyticalTable/defaults/Column/Cell.tsx
+++ b/packages/main/src/components/AnalyticalTable/defaults/Column/Cell.tsx
@@ -1,12 +1,12 @@
 import React from 'react';
 
-export const Cell = ({ cell: { value = '', isGrouped }, column, row, webComponentsReactProperties }) => {
+export const Cell = ({ cell: { value = '', isGrouped }, row, webComponentsReactProperties }) => {
   let cellContent = `${value ?? ''}`;
   if (isGrouped) {
     cellContent += ` (${row.subRows.length})`;
   }
   return (
-    <span title={cellContent} className={webComponentsReactProperties.classes.tableText} data-id={column.id}>
+    <span title={cellContent} className={webComponentsReactProperties.classes.tableText}>
       {cellContent}
     </span>
   );
--- a/packages/main/src/components/AnalyticalTable/defaults/Column/Cell.tsx
+++ b/packages/main/src/components/AnalyticalTable/defaults/Column/Cell.tsx
@@ -1,12 +1,12 @@
 import React from 'react';
 
-export const Cell = ({ cell: { value = '', isGrouped }, row, webComponentsReactProperties }) => {
+export const Cell = ({ cell: { value = '', isGrouped }, column, row, webComponentsReactProperties }) => {
   let cellContent = `${value ?? ''}`;
   if (isGrouped) {
     cellContent += ` (${row.subRows.length})`;
   }
   return (
-    <span title={cellContent} className={webComponentsReactProperties.classes.tableText}>
+    <span title={cellContent} className={webComponentsReactProperties.classes.tableText} data-id={column.id}>
       {cellContent}
     </span>
   );
--- a/packages/main/src/components/AnalyticalTable/defaults/Column/Cell.tsx
+++ b/packages/main/src/components/AnalyticalTable/defaults/Column/Cell.tsx
@@ -1,12 +1,16 @@
 import React from 'react';
 
-export const Cell = ({ cell: { value = '', isGrouped }, row, webComponentsReactProperties }) => {
+export const Cell = ({ cell: { value = '', isGrouped }, column, row, webComponentsReactProperties }) => {
   let cellContent = `${value ?? ''}`;
   if (isGrouped) {
     cellContent += ` (${row.subRows.length})`;
   }
   return (
-    <span title={cellContent} className={webComponentsReactProperties.classes.tableText}>
+    <span
+      title={cellContent}
+      className={webComponentsReactProperties.classes.tableText}
+      data-column-id-cell-text={column.id}
+    >
       {cellContent}
     </span>
   );
--- a/packages/main/src/components/AnalyticalTable/hooks/useAutoResize.tsx
+++ b/packages/main/src/components/AnalyticalTable/hooks/useAutoResize.tsx
@@ -6,37 +6,12 @@ import { CELL_PADDING_PX } from './useDynamicColumnWidths.js';
 const handleAutoResize = (props, { instance }) => {
   return {
     ...props,
-    onDoubleClick: (e, start, end, accessor, onAutoResize) => {
-      // Find first in column to compare to accessor
-      let columnFirst;
-      const { columnOrder } = instance.state;
-      let colOrder = [];
-      if (columnOrder.length) {
-        colOrder = columnOrder.filter(
-          (col) =>
-            col !== '__ui5wcr__internal_selection_column' &&
-            col !== '__ui5wcr__internal_highlight_column' &&
-            col !== '__ui5wcr__internal_navigation_column'
-        );
-        columnFirst = colOrder[0];
-      }
-      if (!columnFirst) {
-        colOrder = instance.visibleColumns?.filter(
-          (col) =>
-            col.id !== '__ui5wcr__internal_selection_column' &&
-            col.id !== '__ui5wcr__internal_highlight_column' &&
-            col.id !== '__ui5wcr__internal_navigation_column'
-        );
-        columnFirst = colOrder[0].id;
-      }
-
-      // Undefined accessor should not occur but checking to be safe
-      columnFirst = columnFirst ? columnFirst === accessor : false;
-
-      const rows = instance.rows;
+    onDoubleClick: (e, start, end, accessor, onAutoResize, isTreeTable, grouped) => {
+      if (!instance.rows.length) return;
       const dispatch = instance.dispatch;
-      const rowSlice = rows.slice(start, end);
-      let largest = getContentPxMax(rowSlice, accessor, instance?.webComponentsReactProperties.uniqueId, columnFirst);
+      let rowSlice = instance.rows.slice(start, end + 1);
+      if (isTreeTable || grouped) rowSlice = getExpandedRowsRecursive(rowSlice);
+      let largest = getMeasureMax(rowSlice, accessor, instance.webComponentsReactProperties.uniqueId);
       largest = largest > DEFAULT_COLUMN_WIDTH ? largest : DEFAULT_COLUMN_WIDTH;
       onAutoResize(
         enrichEventWithDetails(e, {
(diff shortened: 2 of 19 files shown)


packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -655,6 +656,105 @@ const AnalyticalTable = forwardRef<AnalyticalTableDomRef, AnalyticalTablePropTyp
...
+    };
+    // End Helpers
+    let largest = 0;
+    // Currently Including Overscan
+    const items = rowVirtualizer.getVirtualItems();
+    const [start, end] = [items[0].index, items[items.length - 1].index];
+
+    for (let i = start; i < end; i++) {
+      // Use the classname for the span where the text lives AnalyticalTable.module.css.js
+      const collection = document.getElementsByClassName(clsx(classNames.tableText));
```
Lukas742, line 700: Searching for a single element should be sufficient here, also I don't think that you need to do that inside the loop, or am I missing something here?
In general, I believe it's enough to retrieve the canvas font once when a user double-clicks the resizer. Given that only standard cell content (text) is supported, with no custom components, the values should be the same for each cell.

Lukas742 (changes requested):
Hi @Papa-Santo 

Thanks a lot for your contribution! :) 

Below you can find the first findings of my review, I will most probably continue it at the start of next week:

- Since expandable rows are rendering a button/icon (depending on theme) inside a cell, `autoResizeable` isn't working properly for them. Please either exclude grouped tables, tree tables (`isTreeTable: true`) and tables with expandable subcomponents (`renderRowSubComponent` with `subComponentsBehavior` `Expandable` or `IncludeHeightExpandable`), or include the icon/button and the indentation in the calculation. If you decide to exclude these modes, please add a note about it to the property.

- Maybe you can reuse some logic of the [useDynamicColumnWidths](https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/main/src/components/AnalyticalTable/hooks/useDynamicColumnWidths.ts) hook. (E.g. [stringToPx](https://github.com/SAP/ui5-webcomponents-react/blob/412206bbe4e47ad731bc1086d1b8bc7f74ad22fe/packages/main/src/components/AnalyticalTable/hooks/useDynamicColumnWidths.ts#L32)) - This is not a requirement, using `canvas` is fine for me as well, I just wanted to point out that such a thing exists.

packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -655,6 +656,105 @@ const AnalyticalTable = forwardRef<AnalyticalTableDomRef, AnalyticalTablePropTyp
     );
   };
 
+  const handleOnAutoResize = (e, accessor) => {
```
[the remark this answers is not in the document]
Papa-Santo: I have been working on refactoring this as a hook. It now follows the pattern of useDynamicColumnWidths and canvas is removed.

I do need some information from the ColumnHeaderContainer that I cannot get from the instance (or maybe there is a way to tap into the information I need, and I'm missing it). 

My first attempt is to return a function to be fed by the onDoubleClick method of the data-resizer. Like useResizeColumnsConfig, it pushes a click handler (onDoubleClick) into the getResizerProps hook provided. 

Unlike, useResizeColumnsConfig, a local onDoubleClick in ColumnHeaderContainer feeds the necessary information to the hook using column.getResizerProps().onDoubleClick. 

I am not sure how I feel about this in general. getResizerProps seems like the right place to put autoResize, but I also receive information I do not need and do not use.

I am also conflicted on detecting the button/icon. I decided to create an ID in Expandable that can be searched for and measured. Otherwise, it might be possible to do this by using properties and themes to decide which type of button/icon is being displayed. This option felt less future proof.

This does not work on TreeTables and tables with strings in the groupBy array (noted in documentation). 

This commit outdates other comments, so I'm leaving them unanswered for now. 

Thanks for your help. Looking forward to further review.

Lukas742:
Hi @Papa-Santo 

I really like the approach using a plugin hook. 👍 
I still found some things that could be improved and I will create a commit for it, so you can take a look. I think this way it's easier for both of us, as explaining some of the changes probably takes more time than just showing them.

Papa-Santo:
Terrific. Thank you!

packages/main/src/components/AnalyticalTable/defaults/Column/Cell.tsx
```diff
@@ -1,12 +1,12 @@
...
 
-export const Cell = ({ cell: { value = '', isGrouped }, row, webComponentsReactProperties }) => {
+export const Cell = ({ cell: { value = '', isGrouped }, column, row, webComponentsReactProperties }) => {
   let cellContent = `${value ?? ''}`;
   if (isGrouped) {
     cellContent += ` (${row.subRows.length})`;
   }
   return (
-    <span title={cellContent} className={webComponentsReactProperties.classes.tableText}>
+    <span title={cellContent} className={webComponentsReactProperties.classes.tableText} data-id={column.id}>
```
Lukas742, line 9: Same as above.

packages/main/src/components/AnalyticalTable/hooks/useAutoResize.tsx
```diff
@@ -0,0 +1,71 @@
...
+        payload: { [accessor]: largest }
+      });
+    }
+  };
+}
+
+function getMeasureMax(accessor, uniqueId, virtualRowsRange, isRtl) {
+  let maxWidth = 0;
+  const firstRowQuery = `[data-component-name="AnalyticalTableBodyScrollableContainer"][data-react-id="${uniqueId}"] [data-virtual-row-index="${virtualRowsRange.startIndex}"]`;
+  const rowsDOM = document.querySelectorAll(`${firstRowQuery}, ${firstRowQuery} ~ [data-virtual-row-index]`);
```
Lukas742, line 41: Querying only the container that contains all rows is more performant, also the `rowsDOM` array now starts with the first visible row, making it only necessary to iterate through rows that are below the `endIndex` of the virtualizer.

packages/main/src/components/AnalyticalTable/hooks/useAutoResize.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const start = virtualRowsRange.startIndex;
+  const end = virtualRowsRange.endIndex;
+
+  for (let i = 0; i <= end - start; i++) {
```
Lukas742, line 45: The loop only measures rows that are currently visible, if the `querySelector` returns `undefined` it jumps to the next index.

packages/main/src/components/AnalyticalTable/hooks/useStyling.ts
```diff
@@ -102,7 +102,8 @@ const getCellProps = (cellProps, { cell: { column }, instance }) => {
     {
       className,
       style,
-      tabIndex: -1
+      tabIndex: -1,
+      id: `${cellProps.key}-${instance?.webComponentsReactProperties.uniqueId}`
```
Lukas742, line 106: I just realized that this can now be removed again :D

packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -655,6 +664,39 @@ const AnalyticalTable = forwardRef<AnalyticalTableDomRef, AnalyticalTablePropTyp
...
+        return rHeight;
+      },
+      [rHeight, rows, renderRowSubComponent, alwaysShowSubComponent, tableState.subComponentsHeight]
+    ),
+    overscan,
+    measureElement,
+    indexAttribute: 'data-virtual-row-index'
+  });
+  // add range to instance for `useAutoResize` plugin hook
+  tableInstanceRef.current.virtualRowsRange = rowVirtualizer.range;
```
Lukas742, line 698: Here I'm mutating the table instance which is fine, because before a user clicks these values are updated anyway. This way the table doesn't rerender just because the values changed (which they do a lot).

Lukas742 (commented):
Hi @Papa-Santo 

I've pushed my changes. I've also added some comments to explain why I changed some things.

I didn't have time to test these changes carefully, but at least all test cases are green :D 

If you find any bugs or would do something different, please feel free to raise your concerns or adjust them directly.

Papa-Santo:
This is a big improvement. I like how you were able to encapsulate everything in the hook. I was wanting to accomplish that same thing.

I only have a concern about the new measurements. Some of it might be due to my lack of understanding of isRtl.

Here is a bug that came up after these changes (See the behavior of resizing age):

https://github.com/SAP/ui5-webcomponents-react/assets/165276313/e6130be9-cc37-42c1-9c9b-5565b5d4e891

I have a fix that changes code inside of getMeasureMax and allows us to measure other types of elements as well:

https://github.com/SAP/ui5-webcomponents-react/assets/165276313/0a9b3dfd-5dbd-4362-8780-d011f915f916

I have to fix a few tests and check the changed code a bit. I should push either later tonight or tomorrow morning for your review.

Lukas742 (commented):
Hi @Papa-Santo 

thanks again for your input!

I think measuring `children` is the right approach here, so great idea :)
I've refactored the measure function again, so we don't have to rely on styles anymore (except for padding). With this approach no additional logic is necessary to support RTL reading direction. Also, custom content is now partially supported. 

- `querySelectors` don't start at the `document` level now, but on "tabel-level"
- The largest cell is now retrieved by checking text `span` and only then each child of the cell is measured. For tree tables this would lead to wrong values depending on nesting levels, so if the cell containing the expand/collapse button is auto-resized each child of each visible cell is measured.
- The description and type of the public event were adjusted.


Looking forward to your feedback :)

Papa-Santo:
I read through the code and tested for a bit now. It's working great! This works for all the normal use cases I tried and it is more efficient.

It is important to add what you added to the description. I did some ecotic table setups and came across a small regression. 

* We lose support for conditional rendering of elements. I'll post two videos to show that.

With the current measuring:

https://github.com/SAP/ui5-webcomponents-react/assets/165276313/68c59197-2d90-4bfb-a385-2cbd94173831

With the measurements before changes:

https://github.com/SAP/ui5-webcomponents-react/assets/165276313/8a707dcf-cb5d-4aa6-8826-8a3dce2b3844



I really like this part of the solution:

> 
> * For tree tables this would lead to wrong values depending on nesting levels, so if the cell containing the expand/collapse button is auto-resized each child of each visible cell is measured.

If you agree that this solutions efficiency is better than the more costly solution which adds support for custom cells with conditional rendering, this seems ready.

Thanks again!

Lukas742:
> We lose support for conditional rendering of elements. I'll post two videos to show that.

Sorry, I didn't go into enough details for custom cell support. I think in the initial implementation we don't need to support that behavior, since we can't support every possible content passed to a cell anyway. If the need arises to also support basic custom content, we can then check to add a way to allow configuring the measuring method. 

> If you agree that this solutions efficiency is better than the more costly solution which adds support for custom cells with conditional rendering, this seems ready.

Yup, for the time being I'd like to keep this functionality :) I'll ask a colleague to have a quick look at this PR, and if he doesn't find anything we're good to merge. Thank you!

Lukas742 (approved):

MarcusNotheis (approved):

PR #5970, opened 2024-06-25
refactor(enums): harmonize enum names

MarcusNotheis:
BREAKING CHANGE: the `MessageBoxActions` enum has been renamed to `MessageBoxAction`
BREAKING CHANGE: the `MessageBoxTypes` enum has been renamed to `MessageBoxType`
BREAKING CHANGE: the `Themes` enum has been renamed to `Theme` and the deprecated "Belize" (`sap_belize`) theme family has been removed


packages/main/src/enums/Theme.ts
```diff
@@ -1,13 +1,10 @@
...
   sap_fiori_3_hcb = 'sap_fiori_3_hcb',
   sap_fiori_3_hcw = 'sap_fiori_3_hcw',
   sap_horizon = 'sap_horizon',
   sap_horizon_dark = 'sap_horizon_dark',
   sap_horizon_hcb = 'sap_horizon_hcb',
-  sap_horizon_hcw = 'sap_horizon_hcw',
-  sap_belize = 'sap_belize',
-  sap_belize_hcb = 'sap_belize_hcb',
-  sap_belize_hcw = 'sap_belize_hcw'
+  sap_horizon_hcw = 'sap_horizon_hcw'
```
Lukas742, line 9: I've removed the deprecated Belize theme family, but I'm not sure if we should offer a codemod for this. We could set it to the default (`sap_horizon`) but this could be undesired in some cases. Maybe it's better that we leave it up to app developers to decide what theme they'll use as replacement.

Lukas742 (approved):

PR #6022, opened 2024-07-04
fix(AnalyticalTable): don't wrap custom header content in `Text` component

Lukas742:


MarcusNotheis (approved):
I don't think we need a breaking change note here, wrapping a Text in the header doesn't sound right for me in AnalyticalTable, especially because we had `maxLines` set to 1 previously

PR #6222, opened 2024-08-19
feat: register current runtime version in window

### LINKED ISSUE #6210: Identify SAP Web Component Version and React Usage on a User Page

Our product currently is working on automation testing for SAP Web Components and is looking for a way to identify the version of the web components being used on a user page.

Additionally, we'd like to know if there's a way to tell if the SAP SuccessFactors are with UI5 Web Components for React or Angular.

For example, with SAP Fiori, it's possible to retrieve the version by running sap.ui.version in the browser console. Is there a similar way to determine the version for SAP Web Components on a page, such as in SAP SuccessFactors pages?

Tried document.querySelector("meta[name='ui5-shared-resources']").Runtimes on the SuccessFactors page, result returned four versions: 1.23.1, 1.20.0, 1.23.1, and 1.21.2. It's still unclear how each of these versions exactly maps to specific components. Is it correct to assume that 1.23.1 might correspond to the ui5/webcomponents-react version, while 1.20.0 could be the version for @ui5/webcomponents according to the mapping table at https://sap.github.io/ui5-webcomponents-react/v1/?path=/docs/getting-started--docs#add-ui5webcomponents-react-to-an-existing-app?
And when I tried running the same query to confirm my assumption on the SAP UI5 Web Components sample React page https://sap-samples.github.io/ui5-webcomponents-sample-react/, I encountered an error: Uncaught TypeError: Cannot read properties of null (reading 'Runtimes').
Tried on https://sap.github.io/ui5-webcomponents-react/v1/?path=/docs/data-display-list--docs got 1.24.8, while seems shall be 1.29.8.
Tried on https://sap.github.io/ui5-webcomponents/components/Button/ got 2.1.1 seems to be correct.

And it seems that due to current implementation limitations, the only way to determine if the page is using UI5 Web Components for React is by manually searching through the sources in the browser's developer tools for any indication of React-related keywords.

It would be helpful to have a more straightforward and automated way to detect and identify the UI5 Web Components versions in use on a page, including whether they are part of a React or Angular project.

### Additional Context

_No response_

### Organization

OpenText ADM UFT One product team

### Declaration

MarcusNotheis:
This PR adds support for importing the current version via the main entrypoint:
```ts
import { version } from '@ui5/webcomponents-react';
import { version } from '@ui5/webcomponents-react-base';
import { version } from '@ui5/webcomponents-react-charts';
import { version } from '@ui5/webcomponents-react-compat';
```

In addition, the current runtime version is registered once per ThemeProvider in the `window` (`globalThis`):
```ts
console.log(globalThis['@ui5/webcomponents-react'].Runtimes);
[{
    "version": "2.0.0-rc.2",
    "major": 2,
    "minor": 0,
    "patch": 0,
    "prerelease": "rc.2"
}]

```

Closes #6210


packages/main/src/components/ThemeProvider/index.tsx
```diff
@@ -79,6 +91,26 @@ const ThemeProvider: FC<ThemeProviderPropTypes> = (props: ThemeProviderPropTypes
     };
   }, []);
 
+  useEffect(() => {
+    if (_versionInfoInjected) {
+      return () => {
+        // noop
+      };
```
Lukas742, line 98: Why do we need a cleanup function here? Couldn't we just use `return;`?

MarcusNotheis: just for the sake of having a cleanup function, I was under the impression that an effect should either always or never return a cleanup function.

Lukas742 (approved):

PR #6269, opened 2024-08-26
fix: remove json export from root index files

Lukas742:
JSON imports are not compatible with all frameworks/build tools.


MarcusNotheis (approved):
Change is fine, but I think I would mark it as fix

PR #6457, opened 2024-10-07
docs(charts): outline a11y limitations

Lukas742:
This PR also adds a `tsconfig.json` to the `.storybook` folder and fixes a misconfiguration in the root tsconfig.

### DIFF
--- a/.storybook/tsconfig.json
+++ b/.storybook/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "extends": "./tsconfig.base.json",
+  "extends": "./../tsconfig.base.json",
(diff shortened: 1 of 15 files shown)


.storybook/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../tsconfig.json",
```
MarcusNotheis, line 2: I think this should rather extend `tsconfig.base.json`

MarcusNotheis (changes requested):

MarcusNotheis (approved):

PR #6471, opened 2024-10-09
fix(Ui5DomRef - TypeScript): update types

Lukas742:


packages/main/src/types/Ui5DomRef.ts
```diff
@@ -1,16 +1,52 @@
+import type { PropertyValue, SlotValue } from '@ui5/webcomponents-base/UI5ElementMetadata.js';
```
MarcusNotheis, line 1: This import might cause issues in bundlers which are interpreting the `exports` map in package.json.
I would recommend adding the `dist` path segment

MarcusNotheis (changes requested):
Except for the import path fine for me, although I'm not 100% sure on whether we should really expose these low level lifecycle events

MarcusNotheis (approved):

PR #6493, opened 2024-10-15
feat(charts): allow configuring the internal `Legend` component

### LINKED ISSUE #5777: [Charts] Add ability to fully customize chart legends

Currently, we can modify the position and the alignment of the chart legend via `chartConfig`. We can also specify if we want to hide the legend through component prop. But we cannot specify exactly how we want the legend to look. Similar to how we have a 'lineConfig' for the lineCharts exposed, we should have a legendConfig so we can fully utilize the props from the lib. https://recharts.org/en-US/api/Legend/. 

Our designers would like to have something like this:
image

However, this is not easily achievable without having access to some of the props such as the legend `formatter`. If the flexibility of the formatter may still not satisfy the needs of the designers, than we can consider to use the `content` prop to have full control. 

Ability to have full control over the Legend component by exposing a config prop that would pass the to the underlying recharts component

_No response_

### Additional Context

_No response_

### Organization

SF

### Declaration

Lukas742:
Closes #5777

### DIFF
--- a/packages/charts/src/interfaces/IChartBaseProps.ts
+++ b/packages/charts/src/interfaces/IChartBaseProps.ts
@@ -1,6 +1,6 @@
 import type { CommonProps } from '@ui5/webcomponents-react';
 import type { ComponentType, ReactNode } from 'react';
-import type { TooltipProps } from 'recharts';
+import type { LegendProps, TooltipProps } from 'recharts';
 import type { ICartesianChartConfig } from './ICartesianChartConfig.js';
 
 export interface IChartBaseProps<T = ICartesianChartConfig> extends Omit<CommonProps, 'onClick'> {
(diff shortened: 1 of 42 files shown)


packages/charts/src/interfaces/IChartBaseProps.ts
```diff
@@ -88,12 +88,22 @@ export interface IChartBaseProps<T = ICartesianChartConfig> extends Omit<CommonP
...
     /**
      * Number that sets the amount of delay time the chart waits when resizing.
      */
     resizeDebounce?: number;
+    //todo: remove "Omit" once supported
+    /**
+     * Defines the configuration object for the internally used `recharts` Legend component.
+     * You can find all possible configuration properties [here](https://recharts.org/en-US/api/Legend).
+     *
+     * __Note:__ It is possible to override internally used props, so please use with caution!
```
MarcusNotheis, line 104: would `overwrite` be more fitting here? https://stackoverflow.com/questions/8651562/overwrite-or-override#:~:text=If%20you're%20replacing%20an,or%20changing%20the%20thing%20overridden.

MarcusNotheis (approved):

PR #6498, opened 2024-10-16
docs(Table): improve story

Lukas742:


packages/main/src/webComponents/Table/Table.mdx
```diff
@@ -58,27 +67,83 @@ const GrowingTable = () => {
...
-            <TableHeaderCell>
-              <Label>Column 1</Label>
-            </TableHeaderCell>
-            <TableHeaderCell>
-              <Label>Column 2</Label>
-            </TableHeaderCell>
-          </>
-        }
-      >
+      <Table {...args} features={<TableGrowing onLoadMore={onLoadMore} type={TableGrowingMode.Scroll} />}>
```
MarcusNotheis, line 70: I would add the `headerRow` in this example for the sake of completeness. Users who a not familiar with storybook might be confused what `...args` should be

packages/main/src/webComponents/Table/Table.mdx
```diff
@@ -58,27 +67,83 @@ const GrowingTable = () => {
...
+          setMode(e.detail.selectedItems[0].textContent);
+        }}
+      >
+        {Object.values(TableSelectionMode).map((selectionMode) => (
+          <SegmentedButtonItem key={selectionMode} selected={selectionMode === mode}>
+            {selectionMode}
+          </SegmentedButtonItem>
+        ))}
+      </SegmentedButton>
+      <Table {...args} features={<TableSelection mode={mode} />}>
```
MarcusNotheis, line 104: I would add the `headerRow` in this example for the sake of completeness. Users who a not familiar with storybook might be confused what `...args` should be

MarcusNotheis (changes requested):

packages/main/src/webComponents/Table/Table.mdx
Lukas742: I've added the rows and used `{...otherProps}` so it's clear that the table does not only implement the given props. E.g. if they're changed via Storybook.

MarcusNotheis (approved):

PR #6628, opened 2024-11-17
fix(VariantManagement): fix dirty-state and `readOnly` behavior

### LINKED ISSUE #6617: VariantManagament: Missing button to save changes to already saved view
### Describe the bug

It looks like in the variant manager, we don’t have the option to save changes to the already saved view. At the moment, after introducing changes, we can only save it as a new view.
It works correctly from time to time, but there is no clear rule.
The `data-readonly` attribute was not present on the item and still there was no `Save` button. We also tried completely removing `readOnly` property from the variant object passed to the component and still no `Save` button.

### Isolated Example

_No response_

### Reproduction steps

1. Create new variant that is not readonly
2. Change something in the config
3. Try to save (missing `Save` button)

### Expected Behaviour

`Save` button is visible when a variant that is not readonly was modified.

### Screenshots or Videos

image

2.1.1

2.2.0

### Browser

Chrome

### Operating System

MacOS Sequoia 15.1

### Additional Context

Analytical tables used with the VariantManagament are very big, up to almost 300 columns and god-knows-how-many rows.
We keep our variants in the local storage, including `readOnly` field and we confirmed that it is not getting any default value when retrieved.

### Relevant log output

_No response_

### Organization

SAP Fioneer

### Declaration

### LINKED ISSUE #6622: VariantManagament: breaks when array of variants is empty
### Describe the bug

Initially when we load our views we have empty variants array so there is no `<VariantItem>` children and also `selectedVariant = ''`. In this case our app is breaking because of the issue here:
```
// packages/main/src/components/VariantManagement/index.tsx
const showSaveBtn = dirtyState && !Object.prototype.hasOwnProperty.call(selectedVariant, 'readOnly');
```

### Isolated Example

_No response_

### Reproduction steps

_No response_

### Expected Behaviour

Children property is typed as `ReactNode | ReactNode[]` so it should not break with empty array.

### Screenshots or Videos

_No response_

2.4.0

2.4.0

### Browser

Chrome

### Operating System

MacOS Sequoia 15.1

### Additional Context

_No response_

### Relevant log output

_No response_

### Organization

SAP Fioneer

### Declaration

superdyzio:
Fixes #6622
Fixes #6617

## Pull Request Checklist

### DIFF
--- a/packages/main/src/components/VariantManagement/index.tsx
+++ b/packages/main/src/components/VariantManagement/index.tsx
@@ -317,9 +317,7 @@ const VariantManagement = forwardRef<HTMLDivElement, VariantManagementPropTypes>
     }
   }, [safeChildrenWithFavorites]);
 
-  // todo: this applies if `readOnly` is set to `false` as well since the value is read via data-attribute and is therefore a string not a boolean
-  const showSaveBtn =
-    dirtyState && selectedVariant && !Object.prototype.hasOwnProperty.call(selectedVariant, 'readOnly');
+  const showSaveBtn = dirtyState && selectedVariant && !selectedVariant.readOnly;
 
   return (
     <div className={variantManagementClasses} style={style} {...rest} ref={ref}>
--- a/packages/main/src/components/VariantManagement/index.tsx
+++ b/packages/main/src/components/VariantManagement/index.tsx
@@ -318,7 +318,8 @@ const VariantManagement = forwardRef<HTMLDivElement, VariantManagementPropTypes>
   }, [safeChildrenWithFavorites]);
 
   // todo: this applies if `readOnly` is set to `false` as well since the value is read via data-attribute and is therefore a string not a boolean
-  const showSaveBtn = dirtyState && !Object.prototype.hasOwnProperty.call(selectedVariant, 'readOnly');
+  const showSaveBtn =
+    dirtyState && selectedVariant && !Object.prototype.hasOwnProperty.call(selectedVariant, 'readOnly');
 
   return (
     <div className={variantManagementClasses} style={style} {...rest} ref={ref}>


packages/main/src/components/VariantManagement/index.tsx
```diff
@@ -318,7 +318,8 @@ const VariantManagement = forwardRef<HTMLDivElement, VariantManagementPropTypes>
   }, [safeChildrenWithFavorites]);
 
   // todo: this applies if `readOnly` is set to `false` as well since the value is read via data-attribute and is therefore a string not a boolean
```
superdyzio, line 320: btw should this be a `todo`? it seems to me more like a note explaining some context

packages/main/src/components/VariantManagement/index.tsx
```diff
@@ -318,7 +318,8 @@ const VariantManagement = forwardRef<HTMLDivElement, VariantManagementPropTypes>
...
-  const showSaveBtn = dirtyState && !Object.prototype.hasOwnProperty.call(selectedVariant, 'readOnly');
+  const showSaveBtn =
+    dirtyState && selectedVariant && !Object.prototype.hasOwnProperty.call(selectedVariant, 'readOnly');
```
Lukas742, line 322: It turns out that there's another issue with this approach. The `hasOwnProperty` check is only there to workaround React 19 (previously) handling data-attributes differently then React 18 and below. At least this was the case in the release candidate version of React we used when implementing this.
Unfortunately there's a big oversight on our end here, because with this implementation the save button is displayed only if `readOnly` isn't set at all.

Could you please refactor it like this?

```suggestion
  const showSaveBtn =
    dirtyState && selectedVariant && !selectedVariant.readOnly
```

I'll take a deeper look into how React19 treats data-attributes in the latest rc and also improve the test of the `readOnly` behavior in a later PR.

packages/main/src/components/VariantManagement/index.tsx
Lukas742: The comment can be removed as it's not applicable anymore.

Lukas742 (changes requested):

Lukas742 (approved):
LGTM 🚢 
Thanks for the contribution!

The failing test is using an rc version of React 19, so we're good to merge this as React19 isn't stable yet. We'll look into this ASAP.

PR #6650, opened 2024-11-22
chore(StyleStore): import `useSyncExternalStore` from `react`

Lukas742:


MarcusNotheis (changes requested):
Please remove the dependency `use-sync-external-store` from the package.json as well

MarcusNotheis (approved):

PR #6696, opened 2024-12-05
feat: update to UI5 Web Components 2.5.0

MarcusNotheis:


packages/main/src/webComponents/UploadCollection/index.tsx
```diff
@@ -123,7 +111,7 @@ const UploadCollection = withWebComponent<UploadCollectionPropTypes, UploadColle
   ['accessibleName', 'noDataDescription', 'noDataText', 'selectionMode'],
   ['hideDragOverlay'],
   ['header'],
-  ['drop', 'item-delete', 'selection-change']
```
MarcusNotheis, line 126: The `drop` event has been removed, but as the native event is still published I guess that should be fine and not a breaking change.

Lukas742 (approved):

PR #7128, opened 2025-03-24
chore(renovate): create rule for `typescript` dep

Lukas742:


MarcusNotheis (commented):
This rule might be a bit too wide, the `matchPackageNames` will also match packages for e.g. vite or eslint plugins.
Maybe we want to use `matchSourceUrls` instead?

PR #7188, opened 2025-04-04
docs: correct terminology from attributes to properties

Lukas742:


.storybook/components/DomRefTable.tsx
```diff
@@ -74,14 +74,14 @@ export function DomRefTable() {
...
+          <Heading>DOM Properties & Methods</Heading>
           <p>
-            This component exposes public attributes and methods. You can use them directly on the instance of the
+            This component exposes public properties and methods. You can use them directly on the instance of the
             component, e.g. by using React Refs.
           </p>
           <MessageStrip hideCloseButton style={{ marginBlockEnd: '10px' }}>
-            This table is showing <b>additional</b> attributes and methods which are not available as props. <br />
-            All props (without event handlers, children, and slots) are available as attributes on the DOM ref as well.
+            This table is showing <b>additional</b> properties and methods which are not available as props. <br />
```
MarcusNotheis, line 83:
```suggestion
            This table is showing <b>additional</b> DOM properties and methods which are not available as props. <br />
```
I think properties which are not available as props sounds a bit confusing

MarcusNotheis (changes requested):

.storybook/components/DomRefTable.tsx
Lukas742: Good point! I've added "React" to `props` as well. Now it should be clear what the difference there is :)

MarcusNotheis (approved):

PR #7210, opened 2025-04-08
fix(AnalyticalTable): update column widths correctly when grouped or reordered

Lukas742:
Internal Reference: MXP-2983


Lukas742:
> I think the name of the ref is a bit misleading, sounds more like `isInitialized`, but that's not related to that change

Good point. 👍 
I've adjusted the name.

PR #7222, opened 2025-04-11
chore(renovate): create rule for `React` root deps

Lukas742:


Lukas742:
Not required anymore as versions are pinned now.

PR #7308, opened 2025-05-07
feat: update to UI5 Web Components 2.10.0

Lukas742:


packages/main/src/webComponents/Tokenizer/Tokenizer.stories.tsx
```diff
@@ -77,6 +77,7 @@ export const WithLogic = {
         setSelectedCountries(selectedTokens);
       }
     };
+    console.log(selectedCountries);
```
MarcusNotheis, line 80: remove?

packages/main/src/webComponents/ShellBarSearch/index.tsx
MarcusNotheis, line 1: We need to expose this file via the index barrel

MarcusNotheis (changes requested):

MarcusNotheis (approved):

PR #7316, opened 2025-05-08
fix(cypress-commands): make the options optional for clickUi5ListItemByText

Jianrong-Yu:
In the last commit for `cypress-commands/src/commands.ts`, new `options` parameter is added as a mandatory parameter for function `clickUi5ListItemByText` but it's not used.

It should be optional as other functions and the default value should be `{}`.

## Pull Request Checklist

### DIFF
--- a/packages/cypress-commands/src/commands.ts
+++ b/packages/cypress-commands/src/commands.ts
@@ -74,7 +74,7 @@ declare global {
        * cy.get('[ui5-list]').clickUi5ListItemByText("List Item")
        * cy.clickUi5ListItemByText("List Item")
        */
-      clickUi5ListItemByText(text: string, options: Partial<ClickOptions>): Chainable<Element>;
+      clickUi5ListItemByText(text: string, options?: Partial<ClickOptions>): Chainable<Element>;
 
       /**
        * Click on an `ui5-option` of the `ui5-select` component by text.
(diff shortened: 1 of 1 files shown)


packages/cypress-commands/src/commands.ts
```diff
@@ -74,7 +74,7 @@ declare global {
        * cy.get('[ui5-list]').clickUi5ListItemByText("List Item")
        * cy.clickUi5ListItemByText("List Item")
        */
-      clickUi5ListItemByText(text: string, options: Partial<ClickOptions>): Chainable<Element>;
+      clickUi5ListItemByText(text: string, options?: Partial<ClickOptions>): Chainable<Element>;
```
Lukas742, line 77: `{force: true}` is not supported by this custom command:

```suggestion
      clickUi5ListItemByText(text: string, options?: Partial<Omit<ClickOptions, 'force'>>): Chainable<Element>;
```

Could you please also add a note to the jsdoc `@param options` annotation. E.g.:

```
@param [options] ClickOptions (without `force`)
```

Lukas742 (changes requested):
Hi @Jianrong-Yu 

thanks a lot for your contribution, it's really appreciated!

Adding the `options` parameter in the type definitions was an oversight. Initially, I planned to remove it again since `force: true` doesn't work, but we can definitely keep it for the other options. The type just needs to reflect that (please see my comment below).

packages/cypress-commands/src/commands.ts
Jianrong-Yu: Code is updated.
I changed the comment to 
```
* @param {Partial<ClickOptions>} [options] ClickOptions (without `force`)
```
Since storybook is failed to parse the type from the code.
it's not very accurate but I think it should be enough 😅

Lukas742 (approved):

PR #7336, opened 2025-05-13
chore: enable eslint & `prettier:all` for all files

Lukas742:


.prettierignore
```diff
@@ -1,5 +1,11 @@
...
+
 yarn.lock
+.next
+dist
+temp
+.out
 
-/.nx/cache
\ No newline at end of file
+.nx/cache
```
MarcusNotheis, line 10: I guess we can exclude `/.nx` entirely is the folder is generated?

MarcusNotheis (approved):

PR #7369, opened 2025-05-21
chore(eslint): check all relevant files

Lukas742:


package.json
```diff
@@ -19,6 +19,7 @@
     "clean": "tsc --build --clean && tsc --build tsconfig.build.json --clean && rimraf temp .out && lerna run clean",
     "clean:remove-modules": "yarn clean && rimraf node_modules",
     "prettier:all": "prettier --write --config ./prettier.config.js \"**/*\"",
+    "lint:all": "eslint .",
```
MarcusNotheis, line 22: Not sure if we nee a dedicated script, I would rather vote for changing the current `lint` script to lint everything

.github/workflows/main.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         run: yarn install --immutable
 
       - name: run eslint
-        run: yarn lint
+        run: yarn lint:all
```
MarcusNotheis, line 103: same as above

MarcusNotheis (approved):

PR #7544, opened 2025-07-17
docs(Toolbar - compat): outline React19 `StrictMode` limitations

Lukas742:


packages/compat/src/components/Toolbar/index.tsx
```diff
@@ -141,6 +141,8 @@ const OVERFLOW_BUTTON_WIDTH = 36 + 8 + 8; // width + padding end + spacing start
  * __Note:__ The overflow popover is mounted only when the overflow button is displayed, i.e., any child component of the popover will be remounted, when moved into it.
  *
  * __Note:__ To prevent duplicate child `id`s in the DOM, all child `id`s get an `-overflow` suffix. This is especially important when popovers are opened by id.
+ *
+ * __Note:__ When using React 19 with `StrictMode` enabled, overflow behavior may not work reliably.
```
MarcusNotheis, line 145: Maybe we should add here that this only affects the development mode and not production builds

MarcusNotheis (changes requested):

PR #7752, opened 2025-09-11
fix(AnalyticalTable): improve `scaleWidthMode` "Grow" and "Smart" 

### LINKED ISSUE #7790: [AnalyticalTable]: [scale width mode "Grow" issues]
### Describe the bug

Hi!  
I use the `AnalyticalTable` with `scaleWidthMode={"Grow"}` as I want the AnalyticalTable to consider the width of the headers as well as the width of the actual content.

When using this feature with lots of columns, that reserve some amount of space, it's seems to be never possible to get rid of the horizontal scroll bar. But the even more pressing issue is, that as soon as almost all columns are displayed, a "tipping point" is reached that seems to trigger a recalculation of column widths. After that the column sizes have increased, but actually less content is visible.

I created a stackblitz example, I will also attach a video.

Please note, that I intentionally used `scaleWidthModeOptions` in the columns, as this is also the use case in my "real" app.

### Isolated Example

https://stackblitz.com/edit/github-myd8tofb

### Reproduction steps

1. Run the stackblitz example
2. Increase / Decrease the width of the preview container
3. The column widths will "jump" at certain points

### Expected Behaviour

As long as there is not enough space to display all columns, the columns should be rendered with their calculated min width.
When there is enough space to display all columns, _**only then**_  the column sizes should be increased based on the available additional free space. No horizontal scroll bar should be visible.

### Screenshots or Videos

https://github.com/user-attachments/assets/0b43594a-5131-4732-aec6-4e4c3eb216b5

2.14.1

2.14.0

### Browser

Chrome

### Operating System

MacOS

### Additional Context

_No response_

### Relevant log output

```shell

```

### Organization

_No response_

### Declaration

Lukas742:
Fixes #7790

### DIFF
--- a/packages/main/src/components/AnalyticalTable/hooks/useDynamicColumnWidths.ts
+++ b/packages/main/src/components/AnalyticalTable/hooks/useDynamicColumnWidths.ts
@@ -16,12 +16,14 @@ interface IColumnMeta {
   width?: number | undefined;
 }
 
-interface ComputedCSSVarValues {
+interface ComputedCSSValues {
   bodyFontFamily: string;
   bodyFontSize: string;
   headerFontFamily: string;
+  rootFontSize: number;
 }
 
+const FALLBACK_FONT_SIZE = 14;
 const ROW_SAMPLE_SIZE = 20;
 const MAX_WIDTH = 700;
 export const CELL_PADDING_PX = 18; /* padding left and right 0.5rem each (16px) + borders (1px) + buffer (1px) */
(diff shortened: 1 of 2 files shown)


Lukas742:
Hi @Na1k, 

Thanks for the suggestion! I’ll take a look as soon as possible. Currently I’m busy preparing the next minor release, but we’ll put this fix on hold for now until I had the time to check your suggestion.

MarcusNotheis (approved):

PR #7839, opened 2025-10-20
fix(AnalyticalTable): fix vertical scrollbar styles and sync logic

### LINKED ISSUE #7831: [AnalyticalTable]: [Scroll doesn't change position]
### Describe the bug

Hi, 

Hi, I noticed a bug in version 2.15.* with AnalyticalTable component. When I added `visibleRowCountMode="AutoWithEmptyRows"`  the scroll position doesn't change when scrolling with the mouse.

IMPORTANT: 
Scroll doesn't work correctly only when version **Vite** is higher than 6..0. 

Regards
Krzysztof

### Isolated Example

https://stackblitz.com/edit/ui5wcr-at-xezyfxpy?file=src%2FApp.tsx

### Reproduction steps

1.
2.
3.
...

### Expected Behaviour

_No response_

### Screenshots or Videos

_No response_

2.15.*

2.15.*

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Relevant log output

```shell

```

### Organization

_No response_

### Declaration

Lukas742:
All of these issues are caused, because of the new global `scrollbar-color` CSS property.

__Chrome:__ 
- scrollbar track height is not updated correctly in some cases
image

_Here the table is scrolled to the last row_

__Solution:__ force reflow to update scrollbar track height

__Firefox:__
- scrollbar container do not automatically adjust to the scrollbar width, leading to hidden scrollbars if no `width` is set.
- scrollbars do not scale in Firefox -> scrollbar container width would have to be calculated according to scale value.

__Solution:__ Use native scrollbar and add padding in last cells to prevent content from being hidden

__General:__
- Improve scroll performance when dragging scrollbar
- Fix issue with scroll synchronization when updating `data`

___

Fixes #7831

### DIFF
--- a/packages/main/src/components/AnalyticalTable/AnalyticalTable.cy.tsx
+++ b/packages/main/src/components/AnalyticalTable/AnalyticalTable.cy.tsx
@@ -4304,6 +4304,66 @@ describe('AnalyticalTable', () => {
     cy.focused().should('have.text', 'Before');
   });
 
+  it('vertical scroll sync', () => {
+    cy.mount(<AnalyticalTable columns={columns} data={generateMoreData(100)} />);
+
+    cy.get('[data-component-name="AnalyticalTableBody"]').scrollTo(0, 2000).should('have.prop', 'scrollTop', 2000);
+    cy.get('[data-component-name="AnalyticalTableVerticalScrollbar"]').should('have.prop', 'scrollTop', 2000);
+
+    cy.get('[data-component-name="AnalyticalTableVerticalScrollbar"]')
+      .scrollTo(0, 3000)
+      .should('have.prop', 'scrollTop', 3000);
+    cy.get('[data-component-name="AnalyticalTableBody"]').should('have.prop', 'scrollTop', 3000);
+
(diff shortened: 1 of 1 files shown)


packages/main/src/components/AnalyticalTable/index.tsx
```diff
@@ -99,6 +100,8 @@ import {
 } from './util/index.js';
 import { VerticalResizer } from './VerticalResizer.js';
 
+const isFirefox = isFireFoxFn();
```
MarcusNotheis: Could lead to an hydration error in SSR environments. This returns always false on the server, on some clients it might be true -> Hydration Error

packages/main/src/components/AnalyticalTable/scrollbars/VerticalScrollbar.tsx
```diff
@@ -13,10 +15,35 @@ interface VerticalScrollbarProps {
   classNames: ClassNames;
 }
 
+const isChrome = isChromeFn();
```
MarcusNotheis: Could lead to an hydration error in SSR environments. This returns always false on the server, on some clients it might be true -> Hydration Error

MarcusNotheis (approved):

PR #8513, opened 2026-04-24
refactor(charts): remove `@ui5/webcomponents-react` peer dep

Lukas742:


Lukas742:
We will keep `@ui5/webcomponents-react` as peer dependency. I.a. because of localization and theming.

PR #8525, opened 2026-04-28
fix(charts): broaden peer dependency ranges

Lukas742:


MarcusNotheis (approved):

MarcusNotheis:
With this PR merged, we can also get rid of the `charts peer dep` action check

Lukas742:
> With this PR merged, we can also get rid of the `charts peer dep` action check

The compat package peer deps still need to be updated after each minor version release. Although, if the "ui5wc-upgrade" skill does its job as intended, the action check becomes redundant - I'll verify this with the next update.

PR #8748, opened 2026-06-30
chore: avoid Rolldown ReferenceError for `version` export

Lukas742:


MarcusNotheis (approved):

Lukas742:
Rolldown bug, not easily fixable. Only affects Chromatic (could affect the docs SB as well, but doesn't seem to currently)

Issue #8582, opened 2026-05-11
ObjectPage in IconTabBar mode: header collapses when re-selecting a non-first section via selectedSectionId


mbritof:
## Description

When using `ObjectPage` in `IconTabBar` mode with controlled `selectedSectionId`, switching back to a non-first section causes a programmatic scroll that collapses the header — even though the user did not scroll.

## Expected Behavior

Changing `selectedSectionId` to a previously-visited non-first section should render that section's content without scrolling or collapsing the header.

## Actual Behavior

The header collapses because `scrollToSectionById` sets `scrollTop` to a non-zero value based on stale scroll state.

## Root Cause

In `ObjectPage/index.tsx`, the `useIsomorphicLayoutEffect` on `selectedSectionId` change (in IconTabBar mode) does:

1. `setInternalSelectedSectionId(selectedSectionId)` — synchronous state update
2. `setTimeout(fireSelectEvent)` — deferred call to `handleOnSectionSelected`

Inside `handleOnSectionSelected`:

```ts
setInternalSelectedSectionId((prevSelectedSection) => {
  if (prevSelectedSection === newSelectionSectionId) {
    scrollToSection(newSelectionSectionId);
  }
  return newSelectionSectionId;
});
```

Since step 1 already set `internalSelectedSectionId` to the new value, `prev === new` is **always true** when `fireSelectEvent` runs. For non-first sections, `scrollToSection` calls `scrollToSectionById`, which calculates:

```ts
objectPageElement.scrollTop =
  sectionRect.top - objectPageRect.top + objectPageElement.scrollTop - scrollMargin;
```

This produces a non-zero `scrollTop`, which triggers header collapse.

For the **first** section this is harmless because `scrollToSection` does `scrollTo({ top: 0 })` instead.

## Reproduction

```tsx
const [selectedSection, setSelectedSection] = useState("section-1");

<ObjectPage
  mode="IconTabBar"
  selectedSectionId={selectedSection}
  onSelectedSectionChange={(e) => setSelectedSection(e.detail.selectedSectionId)}
  headerArea={<ObjectPageHeader>Header content</ObjectPageHeader>}
  titleArea={<ObjectPageTitle header="Title" />}
>
  <ObjectPageSection id="section-1" titleText="First">Content 1</ObjectPageSection>
  <ObjectPageSection id="section-2" titleText="Second">Content 2</ObjectPageSection>
[...]

PR #8817, opened 2026-07-24
fix(ObjectPage): re-expand header on tab switch in IconTabBar mode

### LINKED ISSUE #8803: ObjectPage in IconTabBar mode — collapsed header cannot be re-expanded by scroll-up after tab switch
### Package name

\`@ui5/webcomponents-react\`

### UI5 Web Components for React version

\`2.24.1\`

### Related

-  #8582 (released in 2.22.1).

### Description

In \`mode=\"IconTabBar\"\`, when the user has already scrolled down to collapse the header and then switches to another tab, the collapsed header can no longer be re-expanded by scrolling up on the newly-selected tab. The user has to scroll **down first**, then back **up**, before the header expands again.

### Reproduction

Live reproduction: https://stackblitz.com/edit/vitejs-vite-gxuja5mx?file=src%2Fmain.tsx

Steps:

1. Render an \`ObjectPage\` with \`mode=\"IconTabBar\"\` and multiple \`ObjectPageSection\`s, each with enough content to make the page scrollable.
2. On the initial tab, scroll down until the header collapses.
3. Click a different tab.
4. Try to scroll up.

**Expected:** the collapsed header expands back into view as the user scrolls up (same as normal scroll-up behavior on a single tab).

**Actual:** page is stuck at \`scrollTop === 0\`. Scrolling up does nothing. User must scroll down first, then up, for the header spacer to reappear and let them re-expand.

### Root cause (from reading the source)

In \`packages/main/src/components/ObjectPage/index.tsx\` (or the compiled \`dist/components/ObjectPage/index.js\` line ~743), the header-content-height spacer div's \`height\` is:

\`\`\`js
(headerCollapsed && !headerPinned || scrolledHeaderExpanded)
  && !toggledCollapsedHeaderWasVisible
  && !(mode === ObjectPageMode.IconTabBar && scrollTimeout.current >= performance.now())
  ? \`\${headerContentHeight}px\`
  : 0
\`\`\`

On section swap the code sets \`scrollTimeout.current = performance.now() + 500\` (see the branch commented *\"Reset scroll for section swap; scrollTimeout preserves current header collapsed/expanded state.\"*). During that 500 ms window the last condition suppresses the spacer.

The fix for #8582 correctly stops the tab-switch-triggered \`scrollToSection\` from expanding the header. But it also removes the collapsed-state spacer from the scroll region, so users can't scroll up to re-expand the header on the new tab — they're forced to scroll down first.

### Suggested fix directions

- Keep the header state preserved (as #8582 intended) **and** keep the scroll spacer available so upward scroll can re-expand it.
- Or expose an opt-out prop so consumers can disable the timeout window for scenarios like this.

Happy to test a patched build.

DougDK:

## Pull Request Checklist

- [x] Reviewed the [Contributing Guidelines](https://github.com/UI5/webcomponents-react/blob/main/CONTRIBUTING.md)
  - Especially the [How to Contribute](https://github.com/UI5/webcomponents-react/blob/main/CONTRIBUTING.md#contribute-code) section
- [x] [Correct commit message style](https://github.com/UI5/webcomponents-react/blob/main/docs/Guidelines.md#commit-message-style)

## Problem

In `IconTabBar` mode, if the user scrolled down far enough to collapse the header and then switched tabs, the header could not be re-expanded. After `scrollTo({ top: 0 })`, `scrollTop` was `0` and `headerCollapsed` was still `true`, leaving the user with no way to scroll further up to trigger re-expansion.

## Solution

Reset `headerCollapsedInternal` to `false` on tab switch so the header expands alongside `scrollTo({ top: 0 })`. The existing `scrollTimeout` guard in `useObserveHeights` already prevents the programmatic scroll from immediately re-collapsing it.

Also removes the redundant `scrollTimeout` clause from the spacer height condition introduced in #8582, which was suppressing scroll room during the tab switch window and had the same root problem.

Fixes #8803

### DIFF
--- a/packages/main/src/components/ObjectPage/ObjectPage.cy.tsx
+++ b/packages/main/src/components/ObjectPage/ObjectPage.cy.tsx
@@ -1931,6 +1931,36 @@ describe('ObjectPage', () => {
     cy.findByText('Custom Header Section Two').should('not.be.visible');
     cy.findByText('Subsection1').should('be.visible');
   });
+
+  it('IconTabBar: header re-expands after tab switch when header was collapsed by scrolling', () => {
+    cy.viewport(1440, 900);
+    cy.mount(
+      <ObjectPage
+        data-testid="op"
+        titleArea={DPTitle}
+        headerArea={DPContent}
+        mode={ObjectPageMode.IconTabBar}
+        style={{ height: '100vh', scrollBehavior: 'auto' }}
+      >
+        <ObjectPageSection key="s1" id="s1" titleText="Section 1">
+          <div style={{ height: '2000px', background: 'lightblue' }}>section 1 content</div>
+        </ObjectPageSection>
+        <ObjectPageSection key="s2" id="s2" titleText="Section 2">
+          <div style={{ height: '2000px', background: 'lightyellow' }}>section 2 content</div>
+        </ObjectPageSection>
+      </ObjectPage>,
+    );
+    cy.wait(50);
+
+    // scroll down far enough to collapse the header
+    cy.findByTestId('op').scrollTo(0, 800);
+    cy.get('[data-component-name="ObjectPageHeaderContent"]').should('not.be.visible');
+
+    // switch tabs — header should re-expand and scroll should reset to top
+    cy.get('[ui5-tabcontainer]').findUi5TabByText('Section 2').click();
+    cy.wait(600);
+    cy.get('[data-component-name="ObjectPageHeaderContent"]').should('be.visible');
+  });
 });
 
 const DPTitle = (
--- a/packages/main/src/components/ObjectPage/index.tsx
+++ b/packages/main/src/components/ObjectPage/index.tsx
@@ -322,9 +322,9 @@ const ObjectPage = forwardRef<ObjectPageDomRef, ObjectPagePropTypes>((props, ref
       }
       return newSelectionSectionId;
     });
-    // Reset scroll for section swap; scrollTimeout preserves current header collapsed/expanded state.
     if (mode === ObjectPageMode.IconTabBar) {
       scrollTimeout.current = performance.now() + 500;
+      setHeaderCollapsedInternal(false);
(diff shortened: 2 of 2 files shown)


Lukas742:
Thanks @DougDK, nice work on this. Your root-cause analysis was solid, and you flagged the `scrollTimeout` clause in the spacer condition, which turned out to matter for the fix too.

Quick process note: per our [contribution guidelines](https://github.com/UI5/webcomponents-react/blob/main/CONTRIBUTING.md#how-to-contribute---the-process), we ask that contributors comment on the issue and get assigned before opening a PR, so we can give guidance up front. That's especially handy on `ObjectPage`, which is more intricate than it looks.

We can't merge it as-is, though. `setHeaderCollapsedInternal(false)` expands the header on every tab switch, which reverses #8582's intent of preserving the collapsed state (and fires `onToggleHeaderArea` as a side effect). If a user deliberately collapsed the header, switching tabs shouldn't re-open it.

The collapse/expand logic has to stay consistent across a lot of paths: pinned/unpinned, Default vs. IconTabBar mode, programmatic `selectedSectionId`/`selectedSubSectionId` changes, subsection navigation, the `scrollTimeout` guard windows, and the intersection observer driving section selection. A fix that works for one often shifts another, which is what's happening here.

The specific snag: "section flush at top" and the scroll-up expand threshold are the same scroll coordinate, so resetting to the top while collapsed always re-expands. We're keeping the header state preserved and landing just past that threshold instead, so scroll-up can still re-expand, with regression tests across both paths.

I've opened #8837 with that approach and added you as co-author, since your repro and the spacer-clause catch fed right into the fix. Going to close this one in favor of that PR. Thanks again for digging in! 🙂
For future contributions, it's best to comment on the issue first so we can flag whether it's a good one to take on. Some fixes we'll might end up handling directly, but the analysis and repro work is always genuinely useful.

Lukas742:
Superseded by #8837

PR #8837, opened 2026-07-31
fix(ObjectPage): allow re-expanding collapsed header after `IconTabBar` tab switch

### LINKED ISSUE #8803: ObjectPage in IconTabBar mode — collapsed header cannot be re-expanded by scroll-up after tab switch

Lukas742:
Fixes #8803


Lukas742:
Converted to draft as the spacer area is currently present after manually collapsing the header when scrolled to top.

